PR #18, opened 2022-09-21
add colab with gradio demo and spaces link on Hugging Face

AK391:
added a colab notebook with a gradio demo and link to web demo on Hugging Face Spaces


Grauen:
Nice Job.

Would be cool if you could also upload a file. Also there should be an option to not trim the audio fo 30 seconds.

jongwook:
Hey AK, now that the demo page is being served at [the openai/whisper space on HF](https://huggingface.co/spaces/openai/whisper), the Gradio notebook doesn't seem necessary here. Please feel free to post at the [show & tell] category about the demo. Thank you!

PR #42, opened 2022-09-22
Avoid keeping redundant copies of model weights in memory during load

drdaxxy:
Before this change, `whisper.load_model` keeps the whole weights file in scope until after tensors are moved to the device.

This might not seem important for sample code, but this change makes it possible to load the `large` model on a free Colab instance, which otherwise runs out of memory (inference afterwards is no problem)!

### DIFF
--- a/whisper/__init__.py
+++ b/whisper/__init__.py
@@ -26,13 +26,14 @@ _MODELS = {
     "large": "https://openaipublic.azureedge.net/main/whisper/models/e4b87e7e0bf463eb8e6956e646f1e277e901512310def2c24bf0e11bd3c28e9a/large.pt",
 }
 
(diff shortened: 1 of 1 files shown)


nlgtuankiet:
I can confirm that it works!

Before: using `large` model on a free Colab instance -> Out of memory error
After: using `large` model on a free Colab instance -> Run successfully!

I'm not sure if these changes affect the output quality, can someone make a comparison to make sure?

jongwook:
This is great! I wasn't 

Let me make some cosmetic changes and then commit.

ShengDong207:
I've successfully reproduced the bug reported in this issue on macOS Ventura. Here are the additional error logs I encountered:

[Please note: I'll need to update this comment with the specific error logs you're seeing]

PR #99, opened 2022-09-24
simplify package installation by removing ffmpeg cli install

fcakyon:
**edit:** you no longer have to accept this PR, just created a new package including this feature: https://github.com/fcakyon/pywhisper

This PR reaplaces `ffmpeg cli tool` + `ffmpeg-python package` with `moviepy package` without any difference in audio waveform values.

This PR removes the external cli tool dependency and simplifies installation since pip install would be enough.

I have tested with `tests/jfk.flac` file and `load_audio()` output is exactly same as before:

```python
dtype: dtype('float32')
max: 0.78271484
min: -0.7235718
shape: (176000,)
size: 176000
```

Also tested with a 1h43min video, works without any issues.

moviepy: https://github.com/Zulko/moviepy

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,25 +17,9 @@ A Transformer sequence-to-sequence model is trained on various speech processing
 
 ## Setup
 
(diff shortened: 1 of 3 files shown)


fcakyon:
you no longer have to accept this PR, just created a new package including this feature: https://github.com/fcakyon/pywhisper

jongwook:
Thanks for the suggestion (and for the other PRs!) moviepy also uses ffmpeg through the command line, indirectly via imageio. Given that we won't need all the image/movie-specific features, ffmpeg-python seems to be a leaner choice.

PR #102, opened 2022-09-24
add srt subtitle export utility

fcakyon:
**edit:** you no longer have to accept this PR, just created a new package including this feature: https://github.com/fcakyon/pywhisper

Currently, transcribed text can be exported only as TXT or VTT.

This PR adds utility for exporting transcription results in SRT subtitle format.

reference: https://en.wikipedia.org/wiki/SubRip


fcakyon:
you no longer have to accept this PR, just created a new package including this feature: https://github.com/fcakyon/pywhisper

thevolcanomanishere:
Nice 😎

jongwook:
Thanks! SRT seems to be in more demand than I anticipated. Simplified a bit and merged :)

PR #126, opened 2022-09-25
Add section for listing community integrations

nn81:
Last pull request had a typo. In the meanwhile added a new extension that was contributed.

The goal is to have a section where the community can find practical usage of whisper for different situations.
Instead of reinventing the wheel, we can check what is already available and either reuse and/or help them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,14 @@ result = whisper.decode(model, mel, options)
 print(result.text)
 ```
 
+## Extensions and integrations
+
+A list of community-contributed extended functionality and integrations with whisper.
+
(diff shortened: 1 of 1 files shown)


jongwook:
Thanks for the suggestion! I don't think I'll have bandwidth to continuously test/endorse/maintain the list of community integrations as they grow, so I'd like to encourage using the [Show and tell](https://github.com/openai/whisper/discussions/categories/show-and-tell) page for advertising applications of Whisper. I think it'd be a great self-organizing list since it has upvotes and bumps the post with the latest activities to the top.

I'll add a section linking to the Show and tell page in the README.md for better visibility.

PR #128, opened 2022-09-25
Context prompt cli option

VulumeCode:
Add --prompt "context" option to lead in the transcription. This can be used to introduce proper names or indicate a technical context.


jongwook:
Thanks! I replaced the walrus operator for Python 3.7 compatibility (which is helpful since Colab is still on py37), and renamed the option `"initial_prompt"`.

nlgtuankiet:
@jongwook that is a great consideration, many people use Colab to explore Whisper. Maintaining compatibility with Colab makes Whisper more accessible.

PR #131, opened 2022-09-25
Accept waveform over STDIN (enable piped input)

lmmx:
The `transcribe.cli()` function currently only takes an `audio_path` which it passes to the `transcribe.transcribe()` function. The `transcribe()` function can take either a file path or a waveform, and the ffmpeg-python simplifies this to just using "pipe:" as the input 'filename'.

This can reduce the unnecessary creation of temporary files as 'intermediate step' from source input and transcript output.

### DIFF
--- a/whisper/transcribe.py
+++ b/whisper/transcribe.py
@@ -6,7 +6,7 @@ from typing import List, Optional, Tuple, Union, TYPE_CHECKING
 import numpy as np
 import torch
 
-from .audio import SAMPLE_RATE, N_FRAMES, HOP_LENGTH, pad_or_trim, log_mel_spectrogram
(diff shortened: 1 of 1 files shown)


jongwook:
I didn't have this in mind when I was writing the code, but just learned that this works:

```bash
cat audio.mp3 | whisper -
```

Thanks to the `ffmpeg` tool interpreting the `-` argument as stdin. There's also [an anecdote](https://github.com/openai/whisper/discussions/135#discussioncomment-3731178) that it works on Powershell as well. So I think this PR is probably not necessary - did you see an error when you passed `-` without the changes in this PR?

lmmx:
Oh, embarrassing, I didn't even check I just got told by a friend that "it doesn't work with piping", you're right! The only difference between my PR and the existing version is that `audio_path` is changed to `stdin` rather than `-` (so you get `-.txt` and `-.vtt`). That's actually perfectly fine, feel free to close in that case :-)

PR #133, opened 2022-09-25
changing the code from using print to using logging

vishnu-anirudh:
Using `logging` instead of `print` makes it easier to log out from various code modules happening at the same time. Also, `print` only logs to `stdout` whereas `logging` logs to the required output.
As a public library, it is better to use `logging` than `print`.

Here is a reference that lists more [reasons ](https://stackoverflow.com/a/6918596) why `logging` is better than `print`.

### DIFF
--- a/whisper/utils.py
+++ b/whisper/utils.py
@@ -1,3 +1,4 @@
+import logging
 import zlib
 from typing import Iterator, TextIO
 
@@ -44,9 +45,9 @@ def format_timestamp(seconds: float):
 
 
 def write_vtt(transcript: Iterator[dict], file: TextIO):
-    print("WEBVTT\n", file=file)
+    logging.info("WEBVTT\n", file=file)
     for segment in transcript:
-        print(
+        logging.info(
             f"{format_timestamp(segment['start'])} --> {format_timestamp(segment['end'])}\n"
             f"{segment['text'].replace('-->', '->')}\n",
             file=file,
--- a/whisper/transcribe.py
+++ b/whisper/transcribe.py
@@ -1,5 +1,6 @@
 import argparse
 import os
+import logging
 import warnings
 from typing import List, Optional, Tuple, Union, TYPE_CHECKING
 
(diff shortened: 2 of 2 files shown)


whisper/utils.py
```diff
@@ -44,9 +45,9 @@ def format_timestamp(seconds: float):
 
 
 def write_vtt(transcript: Iterator[dict], file: TextIO):
-    print("WEBVTT\n", file=file)
+    logging.info("WEBVTT\n", file=file)
     for segment in transcript:
-        print(
+        logging.info(
```
EliEron, line 50: Did you test this code? As far as I'm aware the logging library does not support writing to a file stream in the same way print does. So this code will likely not work.

jongwook:
I am fully aware of why/when `logging` is better than `print()`, but I'd be inclined to keeping `print()` in `transcribe.py` because it is mainly intended to give user feedback and progress updates, and also because it serves as an example script demonstrating a strategy for long-form transcription, rather than a full-fledged application.

Without any configuration, `logging.info()` and `logging.debug()` will do nothing, which would be bad. Requiring users who just want to try out Whisper to configure Python logging would also be a bad UX.

Also, your edits to `write_vtt()` is incorrect as @EliEron pointed out.

vishnu-anirudh:
> I am fully aware of why/when `logging` is better than `print()`, but I'd be inclined to keeping `print()` in `transcribe.py` because it is mainly intended to give user feedback and progress updates, and also because it serves as an example script demonstrating a strategy for long-form transcription, rather than a full-fledged application.
> 
> Without any configuration, `logging.info()` and `logging.debug()` will do nothing, which would be bad. Requiring users who just want to try out Whisper to configure Python logging would also be a bad UX.
> 
> Also, your edits to `write_vtt()` is incorrect as @EliEron pointed out.

@jongwook Apologies I did't know why `print` was still retained. That sounds good.
@EliEron Thanks for your comment. It seems to work at my end. Perhaps I forgot to include the config changes which I have locally. But given that the PR is closed I won't proceed further with it.

PR #140, opened 2022-09-26
fix: transcribe verbosity

bottledmind:
I've hidden "Detected langauge: " massage behind `verbose` flag.
Also fixed the flag usage with tqdm

### DIFF
--- a/whisper/transcribe.py
+++ b/whisper/transcribe.py
@@ -20,7 +20,7 @@ def transcribe(
     model: "Whisper",
     audio: Union[str, np.ndarray, torch.Tensor],
     *,
-    verbose: bool = False,
+    verbose: Optional[bool] = None,
     temperature: Union[float, Tuple[float, ...]] = (0.0, 0.2, 0.4, 0.6, 0.8, 1.0),
     compression_ratio_threshold: Optional[float] = 2.4,
     logprob_threshold: Optional[float] = -1.0,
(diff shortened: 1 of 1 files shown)


jongwook:
I actually intended the two to be displayed even when `--verbose` is False, thinking it'd help the user notice any language detection errors and show the progress/ETA without flooding the console.

bottledmind:
@jongwook well, ETA does not work when verbose=True, which is misleading as it seem to me

jongwook:
The language tag is returned in the result, but for those calling `whisper` in the command line (as it seems to be a popular use case) the detail is lost.

Regarding ETA: that's true; it's not the most intuitively named options but I thought both formats have merits (like below) and thought it'd be nice to have these verbose/concise output options, but probably better not mix up those two.

image

bottledmind:
How about str verbosity like "full", "minimal" and "silent"? I might add it real quick

jongwook:
I'd like to avoid magic string if possible. It's a bit unorthodox but I'm inclined to making it an `Optional[bool]` and use `None`/`False`/`True` so `verbose=None` is the default in `transcribe()` but from the command line it selects either `True` or `False`.

I can make these edits on this PR if you'd like/don't mind.

bottledmind:
Don't bother yourself, I have nothing to do at the moment

whisper/transcribe.py
```diff
@@ -170,7 +172,7 @@ def add_segment(
     num_frames = mel.shape[-1]
     previous_seek_value = seek
 
-    with tqdm.tqdm(total=num_frames, unit='frames', disable=verbose) as pbar:
+    with tqdm.tqdm(total=num_frames, unit='frames', disable=verbose is False) as pbar:
```
jongwook, line 175: Could you make this `verbose is not False`? Thank you.

bottledmind: Done

PR #161, opened 2022-09-28
Update README.md

szpasztor:
Add instruction for installation error: https://github.com/openai/whisper/discussions/160


ExtReMLapin:
When you really want to be in the contributors list

PR #191, opened 2022-09-29
Don't update duration if last timestamp is timestamp_begin

vickianand:
There is bug causing start and end timestamps of the segments to be same when they shouldn't be.
If the only timestamp-token found is the `tokenizer.timestamp_begin` then we shouldn't update the duration (to 0). If there is really no speech, then anyway the segment won't be appended by the `add_segment` function.
Before this fix:
image
After this fix:
image

### DIFF
--- a/whisper/transcribe.py
+++ b/whisper/transcribe.py
@@ -220,7 +220,7 @@ def transcribe(
             else:
                 duration = segment_duration
                 timestamps = tokens[timestamp_tokens.nonzero().flatten()]
-                if len(timestamps) > 0:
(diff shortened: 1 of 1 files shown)


jongwook:
Thanks. I think it's another failure case where Whisper didn't sample any intermediate timestamp tokens. Would you be able to upload/email the audio file?

vickianand:
Here is the audio file: https://drive.google.com/file/d/1WG17h-3U565uWpl_jW2bkO3eS_l73H7f/view?usp=sharing
For the first three 30 sec chunks, it doesn't predict intermediate timestamps when run with:
`--model medium.en --language en  --initial_prompt "um, uh"`

PR #219, opened 2022-10-01
Fix timestamps and strip extraneous whitespace in WebVTT output

tomstuart:
YouTube won’t accept the generated .vtt files if there are any timestamps beyond an hour, because the hour field isn’t formatted correctly. It also shows extraneous whitespace around the segment text because we’re not stripping it. This PR fixes both problems.


Kabanosk (commented):
LGTM!

PR #240, opened 2022-10-04
Fixed CoW RuntimeError in DecodingTask.run()

CorentinJ:
To reproduce:
```
import whisper
import torch

model = whisper.load_model("tiny", device="cuda")
options = whisper.DecodingOptions(without_timestamps=True, fp16=True)
mels = torch.stack([torch.randn(80, 3000, device="cuda") for _ in range(10)])
with torch.no_grad():
    outputs = model.decode(mels, options)
```
Gives
```
tokens[:, self.sot_index + 1] = lang_tokens  # write language tokens
RuntimeError: unsupported operation: more than one element of the written-to tensor refers to a single memory location. Please clone() the tensor before performing the operation.
```

Replacing the expand by a repeat (or calling clone() after expand) solves the issue

### DIFF
--- a/whisper/decoding.py
+++ b/whisper/decoding.py
@@ -615,7 +615,7 @@ class DecodingTask:
         n_audio: int = mel.shape[0]
 
         audio_features: Tensor = self._get_audio_features(mel)  # encoder forward pass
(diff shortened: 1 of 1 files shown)


jongwook:
Thank you!

PR #248, opened 2022-10-05
Update README to include instructions for .en models

mohiitaa:
Using the method `model.transcribe("audio.wav")` with `.en` models (`tiny.en`, `base.en`, `small.en`, `medium.en`) results in the error:
```
ValueError: This model doesn't have language tokens so it can't perform lang id
```

The method needs to be called with an additional argument `language="en"` for `.en` models:
`model.transcribe("audio.mp3", language="en")`

This PR adds this specific instruction to the README for clarity and help to others using these models.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,16 @@ print(result["text"])
 
 Internally, the `transcribe()` method reads the entire file and processes the audio with a sliding 30-second window, performing autoregressive sequence-to-sequence predictions on each window.
 
+The `transcribe()` method requires the argument `language="en"` when using `.en` models:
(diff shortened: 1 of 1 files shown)


jongwook:
Hi, I realized that this is a bit of an annoyance and updated the code (in d18e9ea5dd2ca57c697e8e55f9e654f06ede25d0) so that it won't complain if `language=` is not given for an English-only model. Please try updating the package (`pip uninstall whisper` followed by `pip install git+https://github.com/openai/whisper.git` should work).

PR #278, opened 2022-10-09
Add --threads option to transcribe

eudoxos:
Torch on CPU uses by default number_of_cores/2. This option allows to override this default.


jongwook:
Thank you; but I'd like to recommend using the `MKL_NUM_THREADS`/`OMP_NUM_THREADS` environment variable for simplicity. For example, on bash:


```bash
MKL_NUM_THREADS=4 whisper audio.mp3
```

eudoxos:
Oh, cool, I did not know torch honored `OMP_NUM_THREADS`. I find program argument (documented by `--help`) a bit more user-friendly, and it will work whether torch uses MKL or OpenMP. Feel free to merge the PR or close it without merging, though.

PR #315, opened 2022-10-13
Add package metadata to setup.py

brainwane:
Add project summary, license, etc. to `whisper` project metadata for display when using "pip show" and similar Python package distribution tools.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,8 +7,12 @@ setup(
     name="whisper",
     py_modules=["whisper"],
     version="1.0",
-    description="",
+    description="Robust Speech Recognition via Large-Scale Weak Supervision",
+    readme="README.md",
+    python_requires=">=3.7",
     author="OpenAI",
+    url="https://github.com/openai/whisper",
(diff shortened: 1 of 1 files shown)


jongwook:
This is a great idea but seems like there're some syntax errors. Can you make sure `python setup.py install` runs fine? Would also appreciate if the spaces around `=` are deleted while doing so. Thanks!

brainwane:
Whoops! Thanks for the catch. Fixed. Verified that `python setup.py install` works fine, as does `pip install`. The metadata now shows up as:

~~~
$ pip show whisper
Name: whisper
Version: 1.0
Summary: Robust Speech Recognition via Large-Scale Weak Supervision
Home-page: https://github.com/openai/whisper
Author: OpenAI
Author-email: 
License: MIT
Location: [redacted; location on my machine]
Requires: ffmpeg-python, more-itertools, numpy, torch, tqdm, transformers
Required-by: 
~~~

PR #399, opened 2022-10-22
Disabled '...' from being generated, since it often gets generated

shervinemami:
Whisper often outputs "..." during short pauses in the middle of sentences and also often at the end of sentences & phrases. This is especially problematic when using many short phrases such as during live dictation or live conversations, it keeps inserting "..." roughly 1 or 2 times per sentence! Using the "prompt" above can reduce this but only by steering it significantly, but I imagine  most users don't ever want "..." in their results, so I recommend filtering out "..." completely.

### DIFF
--- a/whisper/tokenizer.py
+++ b/whisper/tokenizer.py
@@ -246,7 +246,7 @@ class Tokenizer:
         keeping basic punctuations like commas, periods, question marks, exclamation points, etc.
         """
         symbols = list("\"#()*+/:;<=>@[\\]^_`{|}~「」『』")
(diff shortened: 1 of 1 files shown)


jpenney:
It might be nice if it were possible to have this set of symbols configurable so users can supply additional symbols to be handled like this, rather than add `...` to the list to exclude. I do want it there, and it can be post processed to remove from the output, but if it's excluded it's lost.

jongwook:
Thanks for the suggestion; there might be cases where `...` could be useful, e.g. when the speaker is hesitating like "I was... I did something". So I'd keep this enabled although I understand it's an opinionated choice.

You can always customize this via the `suppress_tokens` argument as explained in: https://github.com/openai/whisper/discussions/374#discussioncomment-3919526

PR #401, opened 2022-10-23
Update Hebrew language code to he per IANA registry

altryne:
Per [IANA registry](https://www.iana.org/assignments/language-subtag-registry/language-subtag-registry), `iw` was deprecated as the code for Hebrew in 1989 and the preferred code is `he`

The correct subtag: 
```
%%
Type: language
Subtag: he
Description: Hebrew
Added: 2005-10-16
Suppress-Script: Hebr
%%
``` 
And the deprecation
```
%%
Type: language
Subtag: iw
Description: Hebrew
Added: 2005-10-16
Deprecated: 1989-01-01
Preferred-Value: he
Suppress-Script: Hebr
%%
```

### DIFF
--- a/whisper/tokenizer.py
+++ b/whisper/tokenizer.py
@@ -29,6 +29,7 @@ LANGUAGES = {
     "fi": "finnish",
     "vi": "vietnamese",
     "iw": "hebrew",
+    "he": "hebrew",
     "uk": "ukrainian",
     "el": "greek",
     "ms": "malay",
--- a/whisper/tokenizer.py
+++ b/whisper/tokenizer.py
@@ -28,7 +28,6 @@ LANGUAGES = {
     "hi": "hindi",
     "fi": "finnish",
     "vi": "vietnamese",
-    "iw": "hebrew",
     "he": "hebrew",
     "uk": "ukrainian",
     "el": "greek",


whisper/tokenizer.py
```diff
@@ -29,6 +29,7 @@
     "fi": "finnish",
     "vi": "vietnamese",
     "iw": "hebrew",
```
altryne, line 31: left the previous code here for backwards compatibility, but the LANGUAGE_TO_CODE now returns `he` as it should

jongwook:
Thanks for pointing this out. We (carelessly) took the language code from [VoxLingua107](http://bark.phon.ioc.ee/voxlingua107/), which had a few entries that did not match the [ISO-639-1 codes](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes).

However, the dictionary `LANGUAGE` makes uses of the [order-sensitive nature](https://stackoverflow.com/a/39980744) of Python dicts, so this PR will break the languages that comes after Hebrew.

https://github.com/openai/whisper/blob/eff383b27b783e280c089475852ba83f20f64998/whisper/tokenizer.py#L280-L282

I think it wouldn't hurt much to break the backward compatibility and replace `iw` with `he` everywhere, maybe redirecting the input "iw" with "he" where appropriate.

altryne:
@jongwook hey, apologies, I completely missed your comment. 
Agree, and changed to `he` per comment.

PR #532, opened 2022-11-15
Suppress non-timestamp tokens at the begging

jumon:
Related discussion: https://github.com/openai/whisper/discussions/424

Although very rare, Whisper may output a non-timestamp token after the `<transcribe>` token even if it is not in the `without_timestamps` mode. However, the transcribe function assumes that the first decoded token (`sliced_tokens[0]`) is always a timestamp token, and as a result, a starting timestamp could be negative if it is not the case.
See around this line:
https://github.com/openai/whisper/blob/9f70a352f9f8630ab3aa0d06af5cb9532bd8c21d/whisper/transcribe.py#L198

This PR fixes the issue by suppressing non-timestamp tokens as a first generated token.

### DIFF
--- a/whisper/decoding.py
+++ b/whisper/decoding.py
@@ -423,10 +423,14 @@ class ApplyTimestampRules(LogitFilter):
                 else:  # cannot be normal text tokens
                     logits[k, : self.tokenizer.eot] = -np.inf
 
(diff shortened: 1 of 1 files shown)


jongwook:
Thank you so much for identifying the problem! I'll merge this after testing a bit on my end.

PR #627, opened 2022-12-02
Allows Whisper AI to be uploaded to pypi as a package

zackees:
This allows WhisperAI to be used as a python package. I've tested this to ensure that it works.

Note that there is already a `whisper` package which is unrelated. I choose `whisper.ai` as the package name. I'm willing to cede this name to openai with no strings attached - just ask.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,5 @@ __pycache__/
 thumbs.db
 .DS_Store
 .idea
-
+/build
+/dist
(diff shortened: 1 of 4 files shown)


jongwook:
Thanks for the suggestion; we're not planning to use semver or PyPI at the moment, in order to minimize my maintenance burden. You [can still list](https://stackoverflow.com/a/35998253) Whisper in your `requirements.txt` or `setup.py`.

PR #630, opened 2022-12-02
Closing file after reading

paulharter:
Adds a context to close the model file after reading it


jongwook:
This is actually not necessary, as CPython will close the file handle as soon as the reference count becomes zero.

PR #659, opened 2022-12-08
Fix bug where mm is mistakenly replaced with hmm in e.g. 20mm

HennerM:
The English normaliser mistakenly replaces 20mm to 20hmm. In this case "mm" is the unit postfix millimetre.

This was caused by treating 10 as number and thus splitting "10" and "mm". The "mm" token was then further replaced with "hmm" according to the English.json mapping.

Removing "mm" from the mapping shouldn't be a problem, since there is already a condition before that that would remove "mm" words entirely from the input.

"mhm" and "mmm" could probably be removed for the same reason.


jongwook:
Thanks. Those replacers are from the post-processing scripts for the CHiME dataset:

https://github.com/kaldi-asr/kaldi/blob/ae8cbe8858f2a66a9b193c82dbe3b0479364165f/egs/chime5/s5/local/wer_output_filter#L19-L21

but I agree it'd be not very relevant to keep it at this point.

PR #727, opened 2022-12-19
Models are smaller?

Abdurrafey-Siddiqui:
Are the models now smaller in size and can be downloaded faster as well?


Venomous912:
Aslo can you implement this to pywhisper? since i dont need ffmpeg for that

PR #731, opened 2022-12-20
Add `no_speech_threshold` and `logprob_threshold` to DecodingOptions

MatthiasReumann:
## Description

To solve the silence issue one has to adjust the  `no_speech_threshold` and `logprob_threshold` values (see [discussion#29](https://github.com/openai/whisper/discussions/29)). At the moment, this isn't as straightforward as one wishes. This PR tries to make setting these options easier.

#### Currently:
```python 
model = whisper.load_model('tiny')
options = whisper.DecodingOptions().__dict__.copy()
options['no_speech_threshold'] = 0.275
options['logprob_threshold'] = None
result = model.transcribe('/path/to/file', **options, verbose=False)
```

#### After this PR
```python
model = whisper.load_model('tiny')
options = whisper.DecodingOptions(language='en', no_speech_threshold=0.275, logprob_threshold=None)
result = model.transcribe('/path/to/file', **options.__dict__, verbose=False)
```

## Implementation
Add  `no_speech_threshold` and `logprob_threshold` to DecodingOptions data class.

### DIFF
--- a/whisper/decoding.py
+++ b/whisper/decoding.py
@@ -99,6 +99,10 @@ class DecodingOptions:
     # implementation details
     fp16: bool = True  # use fp16 for most of the calculation
 
+    # silence and failure parameters
+    no_speech_threshold: float = 0.6
+    logprob_threshold: Optional[float] = -1.0
+
 
 @dataclass(frozen=True)
 class DecodingResult:


jongwook:
Hi, the class `DecodingOptions` is intended to be used for the [`decode()` function](https://github.com/openai/whisper/blob/main/whisper/decoding.py#L681), which can also be passed via the keyword arguments in `transcribe()`.

Since `no_speech_treshold` and `logprob_threshold` are other arguments to the `transcribe()` function and not used in `decode()`, they shouldn't be in the dataclass `DecodingOptions`.

In your example, you can achieve what you wanted with:

```python
model = whisper.load_model('tiny')
options = whisper.DecodingOptions(language='en')
result = model.transcribe(
    '/path/to/file',
    no_speech_threshold=0.275,
    logprob_threshold=None,
    verbose=False,
    **options.__dict__
)

```

Atefeh197:
> Hi, the class `DecodingOptions` is intended to be used for the [`decode()` function](https://github.com/openai/whisper/blob/main/whisper/decoding.py#L681), which can also be passed via the keyword arguments in `transcribe()`.
> 
> Since `no_speech_treshold` and `logprob_threshold` are other arguments to the `transcribe()` function and not used in `decode()`, they shouldn't be in the dataclass `DecodingOptions`.
> 
> In your example, you can achieve what you wanted with:
> 
> ```python
> model = whisper.load_model('tiny')
> options = whisper.DecodingOptions(language='en')
> result = model.transcribe(
>     '/path/to/file',
>     no_speech_threshold=0.275,
>     logprob_threshold=None,
>     verbose=False,
>     **options.__dict__
> )
> ```

Hi 

What is the meaning of this parameter "no_speech_treshold"?
I mean is there any sources to know about the whisper's parameter?

PR #804, opened 2023-01-03
Add info about WER and which way it goes

mikkovedru:
Without the explanations it is impossible to know what the numbers mean and if the result with a smaller or a bigger number is better.


RijulGupta-DM:
WER stands for Word Error Rate. Lower is better. AWS Transcribe has a WER ~20% in English.

PR #812, opened 2023-01-05
Use ndimage.median_filter instead of signal.medfilter

mu4farooqi:
For a 30s long audio file which didn't have any silence, ndimage.median_filter took 7s where signa.medfilter took 30s.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,26 @@
+name: test
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+jobs:
+  whisper-test:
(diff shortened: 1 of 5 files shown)


mu4farooqi:
I just tried using `cupy.signal.medfilt` and it is faster by at-least 10X. I'll update the CL.

glangford:
> For a 30s long audio file which didn't have any silence, ndimage.median_filter took 7s where signa.medfilter took 30s.

Some scipy history on the two implementations, just FYI

ndimage.median_filter vs signal.medfilt https://github.com/scipy/scipy/issues/9680

doublex:
`cupy` is only used once.
Is it worth the extra dependency?

PR #869, opened 2023-01-20
word-level timestamps in `transcribe()`

jongwook:


ryanheise:
This DTW dependency introduces a licence incompatibility, but an alternative was suggested earlier in the discussions from memory.

whisper/transcribe.py
```diff
@@ -8,15 +8,104 @@
...
+    time_offset = segments[0]["seek"] * HOP_LENGTH / SAMPLE_RATE
+    word_boundaries = np.pad(np.cumsum([len(t) for t in word_tokens]), (1, 0))
+    start_times = time_offset + jump_times[word_boundaries[:-1]]
+    end_times = time_offset + jump_times[word_boundaries[1:]]
+
+    for segment in segments:
+        segment["words"] = []
+
+    for i, (word, start, end) in enumerate(zip(words, start_times, end_times)):
+        if word.startswith("<|") or word.strip() in ".,!?、。":
```
ryanheise, line 96: I think you're missing ？ and others. See:

https://github.com/jianfch/stable-ts/blob/main/stable_whisper/stabilization.py#L440%E2%88%92L443

ryanheise: Would actually be convenient to actually re-insert the punctuation tokens so that concatenating all the words is the same as concatenating all the tokens. That would just make processing easier on the consumer end. For reference, Amazon Transcribe includes timestamped punctuation tokens in the results.

KaiserChr:
Hi!
I tried out this branch with ```kwargs['word_level_timestamps'] = True``` but the model performed very slowly. In addition (or rather due to) it started to hallucinate like mad. 
Im using chunks of short (couple of seconds) audio data in german produced by a VAD for live transcription.

Maybe its a problem on my side, maybe anyone can try to reproduce?

jongwook:
Thanks for the comments, all -- this is work in progress and not quite ready for merging. I'm trying to address both hallucination and performance concerns.

glangford:
Yet another DTW implementation, fyi. Can't vouch for it other than to say that it is Apache licensed, recently updated, has both pure Python and C implementations.

https://github.com/wannesm/dtaidistance

whisper/timing.py
```diff
@@ -0,0 +1,198 @@
...
+        for i, block in enumerate(model.decoder.blocks)
+    ]
+
+    tokens = torch.tensor(
+        [
+            *tokenizer.sot_sequence,
+            tokenizer.timestamp_begin,
+            *[t for segment in segments for t in segment["tokens"]],
+            tokenizer.timestamp_begin + mel.shape[-1] // 2,
+            tokenizer.eot,
```
Jeronymous, line 144: I don't thing these last two tokens are needed to estimate word timestamps.

whisper/timing.py
```diff
@@ -0,0 +1,198 @@
...
+        ]
+    ).to(model.device)
+
+    with torch.no_grad():
+        model(mel.unsqueeze(0), tokens.unsqueeze(0))
+
+    for hook in hooks:
+        hook.remove()
+
+    weights = torch.cat(QKs[-6:])  # layers * heads * tokens * frames
```
Jeronymous, line 154: Why only considering (at most) the last 6 layers?

whisper/timing.py
```diff
@@ -0,0 +1,198 @@
...
+
+    for segment in segments:
+        segment["words"] = []
+
+    for i, (word, start, end) in enumerate(zip(words, start_times, end_times)):
+        if word.startswith("<|") or word.strip() in ".,!?、。":  # TODO: expand
+            continue
+
+        segment = segments[token_sources[word_boundaries[i]]]
+        segment["words"].append(dict(word=word, start=round(start, 2), end=round(end, 2)))
```
Jeronymous, line 192: Would it be possible to add a confidence score based on the average log proba for each words?
This can be a useful feature, available with very little additional computations.

whisper/timing.py
```diff
@@ -0,0 +1,198 @@
...
+    N, M = x.shape
+    cost = np.ones((N + 1, M + 1), dtype=np.float32) * np.inf
+    trace = -np.ones((N + 1, M + 1), dtype=np.float32)
+
+    cost[0, 0] = 0
+    for j in range(1, M + 1):
+        for i in range(1, N + 1):
+            c0 = cost[i - 1, j - 1]
+            c1 = cost[i - 1, j]
+            c2 = cost[i, j - 1]
```
Jeronymous, line 65: Maybe one of these two is not needed, as it doesn't really make sense to attribute a same timestamp to several tokens.
Well... not fully sure. Maybe it's useful when a lot of text has to be aligned with a small portion of audio (which can happen when Whisper "inner language model" is stuck).

Jeronymous (commented):
That feature of getting word-level timestamps is useful for plenty of applications. Excellent work!

Question: Would it be possible to hook attention weights while the model is transcribing, to avoid doing inference twice?
It would be more efficient, and would maybe better take into account factors like the context (conditioning on previous text...).

whisper/transcribe.py
ryanheise: I notice there is now a TODO comment leaning to the first approach:

```python
if word.startswith("<|") or word.strip() in ".,!?、。":  # TODO: expand
```

I'm not sure if you've already committed to that approach, but I would vote for not removing the punctuation, so that whether a consumer wants to traverse the entire content by token, by word, or by segment, they can and do it in either of these 3 ways and in all the content is there (the concatenation of each result is identical). Otherwise if I consume the results by word, I would need to simultaneously look up one of the other two results to cross reference the, and look for the bits that were omitted from the sequence. Here is how [Amazon Transcribe](https://docs.aws.amazon.com/transcribe/latest/dg/how-input.html) does it, for example.

On the other hand, if that is not persuasive, you might consider instead making it an option whether or not to strip out the punctuation.

(I note also that if you just left the punctuation in, the consumer would still have the ability to filter them out.)

whisper/triton_ops.py
```diff
@@ -0,0 +1,92 @@
...
+
+try:
+    import triton
+    import triton.language as tl
+except ImportError:
+    raise RuntimeError("triton import failed; try `pip install --pre triton`")
+
+
+@triton.jit
+def dtw_kernel(cost, trace, x, x_stride, cost_stride, trace_stride, N, M, BLOCK_SIZE: tl.constexpr):
```
taylorchu, line 15: triton jit requires gcc/clang, python-dev, and cuda-dev at runtime. Please consider some lighter-weight alternatives.

vishalsantoshi: +1.

whisper/transcribe.py
jongwook: Thanks for the suggestions; I've updated so that the punctuation characters are attached to the adjacent words, while keeping the word's timestamps.

whisper/timing.py
jongwook: This was a trick to nudge the DTW path to go along these tokens so that the last few words have more accurate timestamp. It's still not perfect, but I settled with using `<|no_timestamps|>` token and no timestamp tokens in the recent commit.

whisper/timing.py
jongwook: This was because the attention weights in the later layers were more indicative of the time alignment. I've updated this part, and now it uses a mask to select which layers and heads to find the alignment.

whisper/timing.py
jongwook: Great point! Added in 5fa4356

whisper/timing.py
jongwook: Yeah it totally makes sense to force a token to have at least one timestamp, which is only about 20 milliseconds. I left this as-is, to handle some failure cases like repetition looping as you mentioned; in the post-processing zero-length segments are removed, and it was usually the case for the generation got stuck on repetition looping.

whisper/triton_ops.py
jongwook: Thanks for the heads up; I made it fall back to pytorch and numba implementations if triton fails with RuntimeError or subprocess.CalledProcessError. I haven't tested this on a non-dev environment, so please feel free to ping me if the fallback does not work for any reason.

whisper/timing.py
Jeronymous: I understand that it's important to have attention weights used to predict the timestamp token for the end of speech segment, but these attention weights are the one you get when the input is the last predicted (sub)word token. I think it's enough. When the input token is the final timestamp the decoder is already focusing on predicting the next thing.
I wonder if things are not shifted by one, because it was a problem I saw with your notebook (the timestamps were assigned to the token before the one it should be).

whisper/timing.py
Jeronymous: Ok interesting. I have to check this masking trick.
I don't understand why later layers are more indicative. Is it an intuition that I am missing, or some empirical results you got from experiments?

whisper/timing.py
Jeronymous: Awesome 👍

whisper/timing.py
vishalsantoshi: We are anecdotally seeing that too, in our tests. The timestamps lag a word for example Got no empirical proof ( very anecdotal ).

ryanheise: I don't see the same observations.

I did notice in earlier commits that the next token after a comma may lag as if the comma was taking up too much time. That seems to have become more accurate in later commits.

whisper/timing.py
vishalsantoshi: We have seen words repeating with same end and start timestamps. Segments were fine. I think we have to move the segment code where you handle duplication ( repetition looping ) to words too.

whisper/timing.py
vishalsantoshi: hmm, we could not come with any empirical evidence either.  May it  was the previous version.

whisper/transcribe.py
ryanheise: Thanks, that looks good. I think if the `prepend_punctuations` and `append_punctuations` parameters were propagated in `transcribe()` and `cli()` that would be quite helpful, since then I could set them to empty strings to emulate the Amazon Transcribe behaviour.

hojinYang:
Hi, thanks for the great work! 

I would like to ask if it is safe to swap to a smaller model (e.g. tiny) for world-level alignment to compute attention scores instead of using the same model (e.g. medium or large ) used to generate transcription. I suspect it could improve performance in terms of inference speed if this option is supported.

whisper/timing.py
```diff
@@ -0,0 +1,305 @@
...
+        logits = model(mel.unsqueeze(0), tokens.unsqueeze(0))[0]
+        token_probs = logits[len(tokenizer.sot_sequence):, :tokenizer.eot].softmax(dim=-1)
+        text_token_probs = token_probs[np.arange(len(text_tokens)), text_tokens].tolist()
+
+    for hook in hooks:
+        hook.remove()
+
+    # heads * tokens * frames
+    weights = torch.stack([QKs[l][h] for l, h in model.alignment_heads.indices().T])
+    weights = weights[:, :, : num_frames // 2]
```
saunair, line 192: Question: Why get weights only in the first `num_frames/2`?

tests/test_timing.py
```diff
@@ -0,0 +1,87 @@
...
+sizes = [
+    (10, 20), (32, 16), (123, 1500), (234, 189),
+]
+shapes = [
+    (10,), (1, 15),  (4, 5, 345), (6, 12, 240, 512),
+]
+
+
+@pytest.mark.parametrize("N, M", sizes)
+def test_dtw(N: int, M: int):
```
saunair, line 18: Was your reason to not use scipy.dtw licensing concerns or just speedup?

tests/test_timing.py
```diff
@@ -0,0 +1,87 @@
...
+    x_cuda = torch.from_numpy(x_numpy).cuda()
+
+    trace_cpu = dtw_cpu(x_numpy)
+    trace_cuda = dtw_cuda(x_cuda)
+
+    assert np.allclose(trace_cpu, trace_cuda)
+
+
+@pytest.mark.parametrize("shape", shapes)
+def test_median_filter(shape):
```
saunair, line 63: Question: Is there a licensing issue using `scipy.median_filter` or is this cuda implementation just faster?

tests/test_timing.py
glangford: dtw-python is GPL, as mentioned here -
https://github.com/openai/whisper/pull/869#issuecomment-1399100852

ryanheise:
I found an interesting edge case with the `small` model where enabling word-level timestamps causes it to repeat the prompt at the end of the audio, while also failing to infer the last word.

```bash
$ ffmpeg -t 29 -i https://audio2.redcircle.com/episodes/6b196013-8672-43d9-be52-4332b3207d93/stream.mp3 test.mp3
$ whisper --model small test.mp3
.../whisper/transcribe.py:98: UserWarning: FP16 is not supported on CPU; using FP32 instead
  warnings.warn("FP16 is not supported on CPU; using FP32 instead")
Detecting language using up to the first 30 seconds. Use `--language` to specify the language
Detected language: English
[00:00.000 --> 00:15.920]  Military veteran Eric Weinstein began 69 Whiskey as a college radio show on 107.7 The
[00:15.920 --> 00:21.720]  Bronx, located on the campus of Ryder University in Lawrenceville, New Jersey.
[00:21.720 --> 00:27.560]  A show once restrained by rules and boundaries now comes straight to you raw, uncensored and
[00:27.560 --> 00:28.960]  unapologetic.

$ whisper --model small --output_format json --word_timestamps True test.mp3
/home/ryan/projects/whisper-clone/whisper/transcribe.py:98: UserWarning: FP16 is not supported on CPU; using FP32 instead
  warnings.warn("FP16 is not supported on CPU; using FP32 instead")
Detecting language using up to the first 30 seconds. Use `--language` to specify the language
Detected language: English
[00:08.040 --> 00:15.940]  Military veteran Eric Weinstein began 69 Whiskey as a college radio show on 107.7 The
[00:15.940 --> 00:21.320]  Bronx, located on the campus of Ryder University in Lawrenceville, New Jersey.
[00:21.720 --> 00:28.980]  A show once restrained by rules and boundaries now comes straight to you raw, uncensored and
[00:28.960 --> 00:28.960]  Military veteran Eric Weinstein began 69 Whiskey as a college radio show on 107.7 The
[00:28.960 --> 00:28.960]  Bronx, located on the campus of Ryder University in Lawrenceville, New Jersey.
[00:28.960 --> 00:28.960]  A show once restrained by rules and boundaries now comes straight to you raw, uncensored and
[00:28.960 --> 00:28.960]  Military veteran Eric Weinstein began 69 Whiskey as a college radio show on 107.7 The
[00:28.960 --> 00:28.960]  Bronx, located on the campus of Ryder University in Lawrenceville, New Jersey.

**Comment by ryanheise** [comment]
I found an interesting edge case with the `small` model where enabling the word-level timestamps option causes it to repeat the prompt at the end of the audio while also failing to infer the last word.

```bash
$ ffmpeg -t 29 -i https://audio2.redcircle.com/episodes/6b196013-8672-43d9-be52-4332b3207d93/stream.mp3 test.mp3

$ whisper --model small test.mp3
.../whisper/transcribe.py:98: UserWarning: FP16 is not supported on CPU; using FP32 instead
  warnings.warn("FP16 is not supported on CPU; using FP32 instead")
Detecting language using up to the first 30 seconds. Use `--language` to specify the language
Detected language: English
[00:00.000 --> 00:15.920]  Military veteran Eric Weinstein began 69 Whiskey as a college radio show on 107.7 The
[00:15.920 --> 00:21.720]  Bronx, located on the campus of Ryder University in Lawrenceville, New Jersey.
[00:21.720 --> 00:27.560]  A show once restrained by rules and boundaries now comes straight to you raw, uncensored and
[00:27.560 --> 00:28.960]  unapologetic.

$ whisper --model small --output_format json --word_timestamps True test.mp3
.../whisper/transcribe.py:98: UserWarning: FP16 is not supported on CPU; using FP32 instead
  warnings.warn("FP16 is not supported on CPU; using FP32 instead")
Detecting language using up to the first 30 seconds. Use `--language` to specify the language
Detected language: English
[00:08.040 --> 00:15.940]  Military veteran Eric Weinstein began 69 Whiskey as a college radio show on 107.7 The
[00:15.940 --> 00:21.320]  Bronx, located on the campus of Ryder University in Lawrenceville, New Jersey.
[00:21.720 --> 00:28.980]  A show once restrained by rules and boundaries now comes straight to you raw, uncensored and
[00:28.960 --> 00:28.960]  Military veteran Eric Weinstein began 69 Whiskey as a college radio show on 107.7 The
[00:28.960 --> 00:28.960]  Bronx, located on the campus of Ryder University in Lawrenceville, New Jersey.
[00:28.960 --> 00:28.960]  A show once restrained by rules and boundaries now comes straight to you raw, uncensored and
[00:28.960 --> 00:28.960]  Military veteran Eric Weinstein began 69 Whiskey as a college radio show on 107.7 The
[00:28.960 --> 00:28.960]  Bronx, located on the campus of Ryder University in Lawrenceville, New Jersey.
```

IgnacioSan22:
Hi @jongwook ,
Since you first release the notebook to obtain word-level timestamps I've been working on this to add to whisper process. And I've tried to test other alingment methods than DTW. Have you tried something else and found out that it works better?

Also, I've been struggling a lot with alucinations, specially for spanish content. I've create a cleaner function at segmet level, is there any smarter way?

ioskevinshah:
is there any chance to have word level timestamps in Whisper API?

jongwook:
Hi @IgnacioSan22, the custom DTW implementation in this PR was for the license issue as noted by others and also for the speed. An alternative is to use the timestamp predictions from the model, but we found that it's less reliable than using the attention patterns like in this PR. If you have solutions using any other algorithms for alignment, please let me know!

The community had some success handling hallucinations by preprocessing the inputs with VAD, like:

- #679
- #397
- https://github.com/m-bain/whisperX

---

Hi @ioskevinshah, this feature is still experimental but we do plan to add it to the API as an option, once we're sure that it's reliable enough.

JeffreyWardman:
@jongwook is there a way to access it via a beta flag for instance? How can we know when something is/isn't added to the API?

jongwook:
For the API, the [speech-to-text guide](https://platform.openai.com/docs/guides/speech-to-text) and the [audio API reference](https://platform.openai.com/docs/api-reference/audio) provide the full documentation of the available features. These documents will be updated accordingly as we roll out new features.

IgnacioSan22:
> Hi @IgnacioSan22, the custom DTW implementation in this PR was for the license issue as noted by others and also for the speed. An alternative is to use the timestamp predictions from the model, but we found that it's less reliable than using the attention patterns like in this PR. If you have solutions using any other algorithms for alignment, please let me know!
> 
> The community had some success handling hallucinations by preprocessing the inputs with VAD, like:
> 
> * [A possible solution to Whisper hallucination #679](https://github.com/openai/whisper/discussions/679)
> * [Whisper WebUI with a VAD for more accurate non-English transcripts (Japanese) #397](https://github.com/openai/whisper/discussions/397)
> * https://github.com/m-bain/whisperX
> 
> Hi @ioskevinshah, this feature is still experimental but we do plan to add it to the API as an option, once we're sure that it's reliable enough.

Hi @jongwook, I've tried the hungarian algorithm and in some cases the results are better, however due to the lack of resources I'm not capable to perform a proper study to find the best alingment algorithm. For hallucinations I've developed a postprocess functions that cleans the segments. It improves quite a lot, but I'll check those references. 

Thanks

glinft:
> For the API, the [speech-to-text guide](https://platform.openai.com/docs/guides/speech-to-text) and the [audio API reference](https://platform.openai.com/docs/api-reference/audio) provide the full documentation of the available features. These documents will be updated accordingly as we roll out new features.

One more question: when will this new feature be rolled out?

whisper/timing.py
```diff
@@ -0,0 +1,305 @@
...
+
+@numba.jit
+def backtrace(trace: np.ndarray):
+    i = trace.shape[0] - 1
+    j = trace.shape[1] - 1
+    trace[0, :] = 2
+    trace[:, 0] = 1
+
+    result = []
+    while i > 0 or j > 0:
```
guillaumekln, line 60: Shouldn't this be a `and` instead of `or`?

ioskevinshah:
> Hi @IgnacioSan22, the custom DTW implementation in this PR was for the license issue as noted by others and also for the speed. An alternative is to use the timestamp predictions from the model, but we found that it's less reliable than using the attention patterns like in this PR. If you have solutions using any other algorithms for alignment, please let me know!
> 
> The community had some success handling hallucinations by preprocessing the inputs with VAD, like:
> 
> * [A possible solution to Whisper hallucination #679](https://github.com/openai/whisper/discussions/679)
> * [Whisper WebUI with a VAD for more accurate non-English transcripts (Japanese) #397](https://github.com/openai/whisper/discussions/397)
> * https://github.com/m-bain/whisperX
> 
> Hi @ioskevinshah, this feature is still experimental but we do plan to add it to the API as an option, once we're sure that it's reliable enough.

any workaround or logic after the API response?

samuelbradshaw:
This is awesome! Is there a way to pass in pre-transcribed text that whisper can use for more accurate alignment?

PR #889, opened 2023-01-24
drop python 3.7 support

jongwook:
(:3 っ)っ


jongwook:
Additional context: Python 3.7 is EOL in 5 months, and we'd like to replace the `transformers` dependency with [tiktoken](https://github.com/openai/tiktoken) for tokenizer, which supports 3.8+.

3.7 supported was added to support Colab, but [Colab has moved to 3.8 now](https://medium.com/google-colab/colab-updated-to-python-3-8-4922f9970a72).

Those on 3.7 can still download the [20230124 release](https://pypi.org/project/openai-whisper/20230124/) which would be last version compatible with 3.7.

PR #914, opened 2023-02-01
Fix infinite loop caused by incorrect timestamp tokens prediction

andrewchernyh:
https://github.com/openai/whisper/discussions/810

### DIFF
--- a/whisper/decoding.py
+++ b/whisper/decoding.py
@@ -412,7 +412,8 @@ class ApplyTimestampRules(LogitFilter):
 
         # timestamps have to appear in pairs, except directly before EOT; mask logits accordingly
         for k in range(tokens.shape[0]):
-            seq = [t for t in tokens[k, self.sample_begin :].tolist()]
+            selected_tokens = tokens[k, self.sample_begin :]
+            seq = [t for t in selected_tokens.tolist()]
             last_was_timestamp = len(seq) >= 1 and seq[-1] >= self.tokenizer.timestamp_begin
             penultimate_was_timestamp = len(seq) < 2 or seq[-2] >= self.tokenizer.timestamp_begin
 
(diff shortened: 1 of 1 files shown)


jongwook:
Thank you!

whisper/decoding.py
```diff
@@ -422,6 +423,11 @@ def apply(self, logits: Tensor, tokens: Tensor):
                 else:  # cannot be normal text tokens
                     logits[k, : self.tokenizer.eot] = -np.inf
 
+            timestamps = sampled_tokens[sampled_tokens.ge(self.tokenizer.timestamp_begin)]
+            if timestamps.numel() > 0:
+                # timestamps shouldn't decrease; forbid timestamp tokens smaller than the last
+                logits[k, self.tokenizer.timestamp_begin : timestamps[-1]] = -np.inf
```
Jeronymous, line 429: This is not enough to prevent the infinite loop (see discussion https://github.com/openai/whisper/discussions/924) because it is not preventing the model to always output `<|0.00|>`

Suggestion:
```
                timestamp_last = max(timestamps[-1], self.tokenizer.timestamp_begin + 1) # Avoid to emit <|0.00|> again
                logits[k, self.tokenizer.timestamp_begin : timestamp_last] = -np.inf
```

Jeronymous (commented):
The infinite loop can still happen, so I suggest to go further with this bugfix

whisper/decoding.py
Jeronymous: I think a better suggestion is:
```
                if last_was_timestamp and not penultimate_was_timestamp:
                    timestamp_last = timestamps[-1]
                else:
                    timestamp_last = timestamps[-1] + 1
                logits[k, self.tokenizer.timestamp_begin : timestamp_last] = -np.inf
```
to force that timestamps are strictly increasing after a speech segment / increasing between the end of a speech segment and the start of the next one.

Is anyone looking at this?

FernanOrtega: Great solution @Jeronymous! I checked it and it works.

With your permission, Im gonna create a new PR to speed up this change. Im gonna mention your suggestion.

PR #973, opened 2023-02-16
Expose punctuation options in cli and transcribe()

ryanheise:
Allows the `prepend_punctuations` and `append_punctuations` options to be set from the CLI or from a python program that calls `transcribe()`.

### DIFF
--- a/whisper/transcribe.py
+++ b/whisper/transcribe.py
@@ -29,6 +29,8 @@ def transcribe(
     condition_on_previous_text: bool = True,
     initial_prompt: Optional[str] = None,
     word_timestamps: bool = False,
+    prepend_punctuations: str = "\"\'“¿([{-",
+    append_punctuations: str = "\"\'.。,，!！?？:：”)]}、",
     **decode_options,
 ):
     """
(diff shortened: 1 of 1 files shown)


jongwook:
Thanks! I'll merge this first and move those magic strings as a global variable somewhere in #869.

ryanheise:
I expected so, and of course there's more of it where that came from ;-)

```python
def transcribe(
    model: "Whisper",
    audio: Union[str, np.ndarray, torch.Tensor],
    *,
    initial_prompt: Optional[str] = None,
    word_timestamps: bool = False,
    prepend_punctuations: str = "\"\'“¿([{-",
    append_punctuations: str = "\"\'.。,，!！?？:：”)]}、",
    **decode_options,
):
```

One of these magic values actually doesn't match up with the default parameter values in `cli()`:

```python
def cli():
    ...
    parser.add_argument("--temperature", type=float, default=0, help="temperature to use for sampling")
```

So when used from the command line, the temperature will default to 0, and via `transcribe()` the temperature will default to `(0.0, 0.2, 0.4, 0.6, 0.8, 1.0)`.

PR #991, opened 2023-02-22
Add per token confidence to each segment.

Pikauba:
As it seems to be useful for some user, I implemented a way to obtain the per token confidence of the model.

The `tokens_probs` will be added to the dictionary of each `segment` the same way `tokens` is. Each `tokens_probs` item's position in the list exactly matches each item's position in the `tokens` list. The confidence of each token can easily be retrieved because the positions of the elements in the two lists match.

As it is difficult to produce a general method to append the tokens which represent a word together depending on the language, I solely focused on producing the token level confidence. I am aware that there is work in progress about a word level timestamp which deals with regrouping tokens per word when possible: [word-level timestamps in transcribe() #869](https://github.com/openai/whisper/pull/869) 

More specifically the [split_tokens_on_spaces function](https://github.com/openai/whisper/blob/8eb29c3ef10559910cbee47b1baedefd8388458a/whisper/tokenizer.py#L285).

So we could eventually use this to merge confidences of the tokens when the language used allows us to do so.

I would like to know if I would be necessary to set the tokens_probs results in each segment as optional and if what are the expectation about the structure of the implementation of it.

### DIFF
--- a/whisper/decoding.py
+++ b/whisper/decoding.py
@@ -106,6 +106,7 @@ class DecodingResult:
     language: str
     language_probs: Optional[Dict[str, float]] = None
     tokens: List[int] = field(default_factory=list)
+    tokens_probs: list[float] = field(default_factory=list)
     text: str = ""
     avg_logprob: float = np.nan
     no_speech_prob: float = np.nan
@@ -196,7 +197,7 @@ class TokenDecoder:
     def reset(self):
         """Initialize any stateful variables for decoding a new sequence"""
 
-    def update(self, tokens: Tensor, logits: Tensor, sum_logprobs: Tensor) -> Tuple[Tensor, bool]:
+    def update(self, tokens: Tensor, logits: Tensor, sum_logprobs: Tensor, tokens_probs: list[list[float]]) -> Tuple[Tensor, list, bool]:
         """Specify how to select the next token, based on the current trace and logits
 
         Parameters
@@ -210,10 +211,16 @@ class TokenDecoder:
         sum_logprobs : Tensor, shape = (n_batch)
             cumulative log probabilities for each sequence
 
+        tokens_probs : list
+            Each confidence of the model's predicted token in the context so far, including the prefix and sot_sequence prob tokens
+            
         Returns
         -------
         tokens : Tensor, shape = (n_batch, current_sequence_length + 1)
             the tokens, appended with the selected next token
+
+        tokens_probs : list
+            confidence of the model's token prediction appended with the selected next token prob
 
         completed : bool
             True if all sequences has reached the end of text
@@ -222,7 +229,7 @@ class TokenDecoder:
         raise NotImplementedError
 
     def finalize(
-        self, tokens: Tensor, sum_logprobs: Tensor
+        self, tokens: Tensor, tokens_probs: list, sum_logprobs: Tensor
     ) -> Tuple[Sequence[Sequence[Tensor]], List[List[float]]]:
         """Finalize search and return the final candidate sequences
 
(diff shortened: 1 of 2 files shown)


ryanheise:
> I am aware that there is work in progress about a word level timestamp which deals with regrouping tokens per word when possible: [word-level timestamps in transcribe() #869](https://github.com/openai/whisper/pull/869)
> 
> More specifically the [split_tokens_on_spaces function](https://github.com/openai/whisper/blob/8eb29c3ef10559910cbee47b1baedefd8388458a/whisper/tokenizer.py#L285).

There is actually word-level confidence in the word-level timestamps branch already ([here](https://github.com/openai/whisper/blob/8eb29c3ef10559910cbee47b1baedefd8388458a/whisper/timing.py#L209)) .

Although maybe I can see some benefit to having a token-level analysis option. For example, German can have [very long compound words](https://theweek.com/articles/463500/8-favorite-ridiculously-long-german-words) where it might be helpful in some use cases to have a finer level of analysis.

Pikauba:
Thank you for pointing that out! I looked at the code and wonder if obtaining the confidence per token once the whole auto-regression process is completed by applying a feed-forward pass with the resulting tokens( as described [here](https://github.com/openai/whisper/blob/8eb29c3ef10559910cbee47b1baedefd8388458a/whisper/timing.py#L182)) and fetch the confidence provided by the model at this point for each token while assuming inference is done without timestamps -> `text_tokens = [token for token in tokens if token < tokenizer.eot]` is equivalent to saving the confidence score per token throught the auto-regression process. (I tested it and it seems like this is actually not the case from what I obtained). 

However, I might be wrong and could misunderstand something about all of it but isn't it right that the timestamp's predictions  are impactful on the predictions of the all the other tokens as they are predicted throught the same inference process?

```python

# coming from line 173 of transcribe.py (https://github.com/openai/whisper/blob/8eb29c3ef10559910cbee47b1baedefd8388458a/whisper/transcribe.py#L173)
tokens = torch.tensor(
    [
        *tokenizer.sot_sequence,
        tokenizer.no_timestamps,
        *text_tokens,
        tokenizer.eot,
    ]
).to(model.device) 

logits = model(mel.unsqueeze(0), tokens.unsqueeze(0))[0]
token_probs = logits[len(tokenizer.sot_sequence):, :tokenizer.eot].softmax(dim=-1)
text_token_probs = token_probs[np.arange(len(text_tokens)), text_tokens].tolist()
```

Pikauba:
Thank you for pointing that out! I looked at the code and wonder if obtaining the confidence per token once the whole auto-regression process is completed by applying a feed-forward pass with the resulting tokens( as described [here](https://github.com/openai/whisper/blob/8eb29c3ef10559910cbee47b1baedefd8388458a/whisper/timing.py#L182)) and fetch the confidence provided by the model at this point for each token while assuming inference is done without timestamps -> `text_tokens = [token for token in tokens if token < tokenizer.eot]` is equivalent to saving the confidence score per token throught the auto-regression process. (I tested it and it seems like this is actually not the case from what I obtained). 

However, I might be wrong and could misunderstand something about all of it but isn't it right that the timestamp's predictions  are impactful on the predictions of the all the other tokens as they are predicted throught the same inference process?

```python

# coming from line 173 of transcribe.py (https://github.com/openai/whisper/blob/8eb29c3ef10559910cbee47b1baedefd8388458a/whisper/transcribe.py#L173)
tokens = torch.tensor(
    [
        *tokenizer.sot_sequence,
        tokenizer.no_timestamps,
        *text_tokens,
        tokenizer.eot,
    ]
).to(model.device) 

logits = model(mel.unsqueeze(0), tokens.unsqueeze(0))[0]
token_probs = logits[len(tokenizer.sot_sequence):, :tokenizer.eot].softmax(dim=-1)
text_token_probs = token_probs[np.arange(len(text_tokens)), text_tokens].tolist()
```

PR #1012, opened 2023-03-01
add task for transliteration to English

spgoswami1:


jongwook:
Simply adding a task token won't work because the model is not trained for it. Transliteration may work with prompting to a limited degree

PR #1021, opened 2023-03-03
remove auxiliary audio extension

ain-soph:
Old:
`'./abc.wav' -> './abc.wav.srt'`

New:
`'./abc.wav' -> './abc.srt'`


jongwook:
Thanks. This is a breaking change but I think it makes sense and will be more compatible to more tools that expect the same basename and different extension

PR #1040, opened 2023-03-07
add always_use_initial_prompt

mercury233:
always_use_initial_prompt: bool
        if True, the initial_prompt will be used to all windows, and condition_on_previous_text
        will be ignored. Enabling this may make the text more consistent if the audio is long
        and you set the initial_prompt properly.

### DIFF
--- a/whisper/transcribe.py
+++ b/whisper/transcribe.py
@@ -540,7 +540,7 @@ def cli():
     parser.add_argument("--suppress_tokens", type=str, default="-1", help="comma-separated list of token ids to suppress during sampling; '-1' will suppress most special characters except common punctuations")
     parser.add_argument("--initial_prompt", type=str, default=None, help="optional text to provide as a prompt for the first window.")
     parser.add_argument("--condition_on_previous_text", type=str2bool, default=True, help="if True, provide the previous output of the model as a prompt for the next window; disabling may make the text inconsistent across windows, but the model becomes less prone to getting stuck in a failure loop")
-    parser.add_argument("--always_use_initial_prompt", type=str2bool, default=False, help="if True, the initial_prompt will be used to all windows, and condition_on_previous_text will be ignored. Enabling this may make the text more consistent if if the audio is long and you set the initial_prompt properly.")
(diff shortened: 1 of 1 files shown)


ryanheise:
I think some variation on this idea might help it to remember your prompting in long audio, but when a window boundary occurs mid sentence, I think it's also important to have the previous window as the prompt.

As a compromise, have you thought about truncating the previous text at a sentence boundary and then prepending the initial prompt before that? It might be the best of both worlds.

mercury233:
> I think some variation on this idea might help it to remember your prompting in long audio, but when a window boundary occurs mid sentence, I think it's also important to have the previous text as the prompt.
> 
> As a compromise, have you thought about truncating the previous text at a sentence boundary and then prepending the initial prompt before that? It might be the best of both worlds.

I agree, but I don't know how to do that.

umar009ali:
I use whisper ai for audio transcription and translation but since 25 February  it is not transcribing and Translating clearly !!!               I would be happy if anybody help me. plz

radurevutchi:
@ryanheise note that this code in `decoding.py:594` truncates the list of all prompt tokens (from the beginning), not the end. That means that simply prepending without checking for prompt window length will not always work.

`tokens = ( [self.tokenizer.sot_prev] + prompt_tokens[-(self.n_ctx // 2 - 1) :] + tokens )`

Paxosman (approved):
Check your write on 22

FurkanGozukara:
testing this right now

FurkanGozukara:
the output after this is just amazing

i dont get why this is still not implemented

FurkanGozukara:
> ```python
>  not condition_on_previous_text or result.temperature > 0.5
> ```

can you share modified file like this? i would like to test. currently it is having problems

mercury233:
> @mercury233 it is hallucinating significantly after this change. anyway to prevent it? 

Sorry, I didn't

jonathanjfshaw:
I have used the same basic idea of applying the initial prompt to every window to supply a dictionary of obscure words that might be in the transcript. It's very effective at boosting recognition of some words. However, I don't see it as in opposition to condition_on_previous text; the basic idea of using context from the end of the previous window to influence understanding of the beginning of the next window is still valuable.

PR #1044, opened 2023-03-07
Use tiktoken

jongwook:
Using [tiktoken](https://github.com/openai/tiktoken) to replace HuggingFace Tokenizers allows faster tokenization and removing tensorflow as a transitive dependency.

A downside is that [tiktoken does not yet provide aarch64 linux wheels](https://github.com/openai/tiktoken/issues/23) while [tokenizers is built even for ppc64le and s390x](https://pypi.org/project/tokenizers/#files). So it may be a blocker for some users..


whisper/tokenizer.py
```diff
@@ -127,21 +129,47 @@
 }
 
 
-@dataclass(frozen=True)
+@dataclass
 class Tokenizer:
     """A thin wrapper around `GPT2TokenizerFast` providing quick access to special tokens"""
```
Majdoddin, line 134:
```suggestion
    """A thin wrapper around `tiktoken` providing quick access to special tokens"""
```

whisper/tokenizer.py
```diff
@@ -127,21 +129,47 @@
...
+        return self.encoding.encode(text, **kwargs)
 
     def decode(
         self, token_ids: Union[int, List[int], np.ndarray, torch.Tensor], **kwargs
     ):
-        return self.tokenizer.decode(token_ids, **kwargs)
+        if isinstance(token_ids, int):
+            token_ids = [token_ids]
+        if isinstance(token_ids, (torch.Tensor, np.ndarray)):
+            token_ids = token_ids.tolist()
```
Majdoddin, line 170:
```suggestion
```

Majdoddin: All calls to decode() in the code pass a List[int] argument. Moreover, callers can always care to pass a list. 
These "if"-tests cost time which spoil the superior runtime of tiktoken.
If the tests should remain then use 
```
if isinstance(token_ids, list):
    pass
elif isinstance(token_ids, int):
    token_ids = [token_ids]
elif isinstance(token_ids, (torch.Tensor, np.ndarray)):
    token_ids = token_ids.tolist()
```
 This makes just one test if the argument is List[int]

Majdoddin: regarding [ `decode_with_timestamps()`](https://github.com/openai/whisper/pull/1044/files#diff-c4ef7e42e8edd6ba1a400cc0fdcb54fccd47a4629979fe66108285a5c3a58053R174) in `tokenizer.py`, why not add the timestampt strings to the tokenizer as special tokens? sth like 
"<|0.00|>", self.timestamp_begin
....
"<|30.00|>", self.timestamp_begin+1500

whisper/tokenizer.py
Majdoddin: I have also implemented whisper with tiktoken, but you have already released yours :) So I make just a few comments.

whisper/tokenizer.py
jongwook: Great points; reflected in the commits below.

petterreinholdtsen:
I am aware of a university running whisper on powerpc, at least, so their upgrade path will be blocked until tiktoken supports more architectures.

whisper/tokenizer.py
```diff
@@ -318,12 +313,17 @@ def split_tokens_on_spaces(self, tokens: List[int]):
...
 @lru_cache(maxsize=None)
-def build_tokenizer(name: str = "gpt2"):
-    os.environ["TOKENIZERS_PARALLELISM"] = "false"
-    path = os.path.join(os.path.dirname(__file__), "assets", name)
-    tokenizer = GPT2TokenizerFast.from_pretrained(path)
+def get_encoding(name: str = "gpt2"):
+    vocab_path = os.path.join(os.path.dirname(__file__), "assets", f"{name}.tiktoken")
+    ranks = {
+        base64.b64decode(token): int(rank)
+        for token, rank in (line.split() for line in open(vocab_path) if line)
```
wsysuper, line 320: Here, you opened the file and left a unclosed handler.

PR #1052, opened 2023-03-07
attempt to fix the repetition/hallucination issue identified in #1046

jongwook:


ryanheise:
Hi @jongwook Not sure if you saw the comment below, but it includes a reproduction case which might be useful:

https://github.com/openai/whisper/pull/869#issuecomment-1445024980

The repetition persists with this PR.

jongwook:
@ryanheise thanks! will look into it...

jongwook:
Thanks all! The incorrect zero-padding of Mel spectrograms as identified in #730 and #838 was contributing to this error. The fix in 477f0be appears to fix the repetition issue.

ryanheise:
> The fix in https://github.com/openai/whisper/commit/477f0befc76456c37b2d2f484095f803592c90fa appears to fix the repetition issue.

I can confirm this fixed my example, thanks! :+1:

> Below is the mp3 fragment converted to video to show the English subtitles.

@glangford FYI the subtitles didn't show in your video.

glangford:
@ryanheise Inline, (on Mac at least) you may need to click on the >> on the right to turn on subtitles. Or download and view with VLC, Quicktime, or whatever and enable subtitles in the viewer.

ryanheise:
Ah, I see. You can also do hard subs this way https://github.com/openai/whisper/discussions/435#discussioncomment-4238872

> using either `small.en` or `medium.en` looks ok although the timing of the last few words is off.

Here is the `base` model for comparison, which appears more accurate on the last few words:


https://user-images.githubusercontent.com/19899190/223701437-fe409ade-c889-4ba5-bd9d-99f5bd869e6b.mp4

glangford:
> Btw have you guys tried with longer audio, e.g. 5 mins long? I am still getting a lot of repetition even with this fix.

I am testing a longer audio now (running on CPU, larger model, transcript+transcribe so it is taking a while). For clarity,
* are you running the 20230307 release version? with, or without `--word_timestamps`?
* the repetitions from "AimeeMullins_2009P.wav" above, are they from verbose print to the console? 

It seems like there are different possible sources of error, in all the different discussions
* model hallucination
* new repetition introduced or magnified by `--word_timestamps True`
* (hand waving) segmentation issues

glangford:
@jongwook Note from @m-bain example above the repetition occurring with verbose print. The repetitions in this example are all "instantaneous" ; eg same start and end time
> [02:14.720 --> 02:14.720] and needless to say, thank God, I wasn't using a thesaurus back then.

they are printed but then immediately cleared by this code, which looks like a bug unique to `--verbose True`

https://github.com/openai/whisper/blob/aac47c98349b98cec5ca7b1be53960fb59f4436b/whisper/transcribe.py#L345

glangford:
@m-bain Given this could you maybe rerun and see if the formal output formats are messed up or not, using `--verbose False`?

m-bain:
This is not a verbose error, and the start times and end times of repetition are not always instantaneous, see output for the .srt file without verbose:

271
00:02:14,440 --> 00:02:14,720
and needless to say, thank God, I wasn't using a thesaurus back<u> then.</u>

272
00:02:14,720 --> 00:02:14,720


273
00:02:15,460 --> 00:02:16,180
<u>I</u> mean from this entry, it would seem that

274
00:02:16,180 --> 00:02:16,360
I<u> mean</u> from this entry, it would seem that

275
00:02:16,360 --> 00:02:16,960
I mean<u> from</u> this entry, it would seem that

276
00:02:16,960 --> 00:02:17,220
I mean from<u> this</u> entry, it would seem that

277
00:02:17,220 --> 00:02:17,620
I mean from this<u> entry,</u> it would seem that

278
00:02:17,620 --> 00:02:17,800
I mean from this entry, it would seem that
>

glangford:
So there are at least two problems then
* verbose mode can print cleared segments
* something else triggered by word_timestamps

Given how close the start/end times are it feels like something related to `seek_shift` is still off
https://github.com/openai/whisper/blob/aac47c98349b98cec5ca7b1be53960fb59f4436b/whisper/transcribe.py#L337

@m-bain Do the same repetitions happen with `word_timestamps False` or no?

m-bain:
Update, I realise there is some specific underline formatting in the word_timestamps, was able to get it working in the end. See here for comparison on word-level timestamp accuracy

image

@jongwook could you share the evaluation for long-form transcription WER? I am unable to reproduce whisper results, right now I report in the vanilla setting -- greedy/beam5 decoding without the heuristic tricks

PR #1076, opened 2023-03-11
Fix github language stats dominated by jupyter notebook

akashmjn:
Tiny PR identical to this :) https://github.com/karpathy/nanoGPT/pull/115 

Adds a `.gitattributes` file telling github to treat jupyter notebooks as "generated code" so that repo is no longer counted as 98% jupyter


glangford (commented):
Alternately use `linguist-vendored` to preserve diffs

glangford:
According to the github doc on overrides, the `linguist-generated` attribute causes files to be "excluded from stats, hidden in diffs". Whereas the `linguist-vendored` attribute causes only exclusion from stats. I think the diffs for notebooks would be useful, so should this be changed use `linguist-vendored` instead?

Reference
https://github.com/github/linguist/blob/master/docs/overrides.md

jongwook:
I think no diffs might actually be better, because edits usually contain base64-encoded images and becomes quite unwieldy on native GitHub diffs, without third-party tools.

PR #1119, opened 2023-03-19
Per Token Confidence + Color terminal example

SinanAkkoyun:
Hello!
I implemented per-token confidence scores and also added a little example under examples/confidence_per_token.py, where you get a fancy colored text output resembling the confidence score of each token:
image
Example WAV: https://www.voiptroubleshooter.com/open_speech/american/OSR_us_000_0010_8k.wav

I am aware of the work of others to this topic, but I as I needed the  I might give the PR a shot.

> (the README.md is identical, I just initially added some lines locally but replaced it again with the full original readme)
> Works with GreedyDecoder for now

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ thumbs.db
 .DS_Store
 .idea
 
+.venv/
+
+samples/
--- a/README.md
+++ b/README.md
@@ -1,3 +1,17 @@
+# Sinan:
+
+Install:
+
+    pip install git+https://github.com/SinanAkkoyun/openai-whisper.git
+
+To update the package to the latest version of this repository, please run:
+
+    pip install --upgrade --no-deps --force-reinstall git+https://github.com/SinanAkkoyun/openai-whisper.git
+
+
+
+
+# OpenAI:
 # Whisper
 
 [[Blog]](https://openai.com/blog/whisper)
--- a/README.md
+++ b/README.md
@@ -1,17 +1,3 @@
-# Sinan:
-
-Install:
-
-    pip install git+https://github.com/SinanAkkoyun/openai-whisper.git
-
-To update the package to the latest version of this repository, please run:
-
-    pip install --upgrade --no-deps --force-reinstall git+https://github.com/SinanAkkoyun/openai-whisper.git
-
-
-
-
-# OpenAI:
 # Whisper
 
 [[Blog]](https://openai.com/blog/whisper)
--- a/examples/confidence_per_token.py
+++ b/examples/confidence_per_token.py
@@ -1,0 +1,57 @@
+# IMPORTANT: This is just for using the local whisper dir as the package directly. Delete until next comment when just installing whisper normally.
+import sys
+from pathlib import Path
+sys.path.insert(0, str(Path(__file__).resolve().parents[1]))
+# end of dev import
+import whisper
+
+import colorsys
+from typing import List
+from whisper.tokenizer import get_tokenizer
+from colorama import init, Style
+
+
+print('Loading model')
+model = whisper.load_model("large")
+
+
+print('Loading audio') # load audio and pad/trim it to fit 30 seconds
+audio = whisper.load_audio("samples/your_audio.wav")
+audio = whisper.pad_or_trim(audio)
+
+
+mel = whisper.log_mel_spectrogram(audio).to(model.device) # make log-Mel spectrogram and move to the same device as the model
+
+
+detect_lang = False
+language = "en"
+if detect_lang: # detect the spoken language
+    print('Detecting language')
+    _, probs = model.detect_language(mel)
+    print(f"Detected language: {max(probs, key=probs.get)}")
+    language=max(probs, key=probs.get)
+
+
+print('Decoding audio') # decode the audio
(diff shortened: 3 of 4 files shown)


SinanAkkoyun:
Current problem: When using prompt, all scores plummit to 0. I don't know why that's the case and help would be really appreciated

SinanAkkoyun:
Keep in mind when supplying prompt: The prompt tokens will also be converted to probs (all prompt tokens result in 0)!

I will update my example to work with prompt input very soon :)

SinanAkkoyun:
TODO: correct propability display when supplying prompts (prompt tokens seem to get assigned prob of 0, if anyone can please help, I'd appreciate it)

SinanAkkoyun:
Fixed probs offset (and prompt offset)

New (correct) output:
image

daxaxelrod:
+1

turnkit:
This can be really useful for proofing the output via something like Subtitle Edit.  

Would really need an command line option to output an additional subtitle though, right?  

I get the impression @jongwook doesn't want to stuff too many features in though, so how does such a useful feature get added without having a fork?

ryanheise:
> I get the impression @jongwook doesn't want to stuff too many features in

Although the colour terminal stuff might be questionable, I think adding per-token timestamps and confidence to the raw JSON results could itself be useful to a wider range of use cases. Exposing the raw data from Whisper would then make it possible to write your own external script on top of that to do the colour terminal staff.

One example of the wider potential uses of per-token data is that German is known to have very long compound words, and if you wanted to break them up (e.g. to compute a line break timestamp, or for sub-word highlighting, etc.), it would be helpful to have access to the raw data per token.

jongwook:
I'm hesitant to add this because the incremental utility of this compared to the probabilities returned by `word_timestamps=True` is quite niche, versus the added complexity & latency due to the additional GPU operations. The decoding logic is already taking as much as the forward pass, and I'm hoping to reduce this overhead. The subword token probabilities are not very useful anyway, because it's usually influenced more by language modeling than from speech recognition.

For the case you need per-token probs, you can add another forward pass without modifying decoding.py (similar to how it's done in [timing.py](https://github.com/openai/whisper/blob/76c901ab8d4558992c44138479c4d69eb52fadcb/whisper/timing.py#L197)) without incurring too much additional latency. It may even be faster than adding GPU operations for every autoregressive step.

The example script looks nifty, but i'd prefer it in the [show and tell](https://github.com/openai/whisper/discussions/categories/show-and-tell) section.

SinanAkkoyun:
I see, thank you for the comprehensive response!

Rtut654:
@SinanAkkoyun thanks for your contribution. Not sure, but seems it works incorrect, 
I made distorted speech example https://drive.google.com/file/d/12zGWllJg6edftcnwuHX_ZHMuwk7PlVjg/view?usp=sharing .
If I don't set the language of decoding,  the output is correct in terms of locating mispronounce (I can read this slavic) though it translates it to random language.
Screenshot from 2023-08-09 10-58-49

But if I set 'en' for decoding the picture is wrong.
Screenshot from 2023-08-09 10-58-53

SinanAkkoyun:
@Rtut654 Hi, I don't quite understand the issue you are having, the "I like to play badminton and football." seems to be correct, the football especially sounds vague in the audio you provided. Could you please tell me more about your issue?

Despite that, the PR is not going to get merged, so I stopped working on it and use that modification in my own work which does not include translation

If the random translation is the problem you are referring to, I believe that my PR did not modify nor change the output prediction by any means, it just grabbed the logits and displays them as confidence

Rtut654:
@SinanAkkoyun 
The issue is in the accuracy of token_probs. The first version (with translation to Ukrainian) gives very accurate result since "like" was also mispronounced very much. Also the word "football " was mispronounced in the last part which is very correctly shown in the first picture. 

I did the same test with other audios setting language of decoding to English.  The picture was same. Somehow it is lowering the prob of the last word even when it is pronounced correctly. At the same time probs of mispronounciations was high. So something is wrong in the way it predicts probs for english.

Rtut654:
Hello, @jongwook ! Could you please look at the issue?

SinanAkkoyun:
@Rtut654 Ah I see, I am sorry but in my testing I was not able to reproduce that behaviour, for me it seemed very fine in regards to unclear spoken words and clearly spoken words, even in german transcriptions.
To me, the football sounded unclearer than the rest, my code does not do much except that it takes the unmodified logits. Maybe the model has a different sense of misspronunciation than you? It has been trained on many accents.

I currently am working on other projects, you can take a look at the file changes yourself and work your way through, if you have a question feel free to ask and I will give my best to clarify

EgSam:
> Hello! I implemented per-token confidence scores and also added a little example under examples/confidence_per_token.py, where you get a fancy colored text output resembling the confidence score of each token:

I think you will need to  amend BeamSearchDecoder as well with the token _probs, otherwise you will face some issues when callin model.transcribe

edsu:
In case it's of interest, I created a small web component to view the Whisper JSON file when `--word_timestamps` has been used. Ideas for improving it would be welcome!

https://edsu.github.io/whisper-transcript/

PR #1135, opened 2023-03-21
IndexError: arrays used as indices must be of integer type

doublex:
Happens on some audio chunks:

 ````
  File whisper/transcribe.py", line 316, in transcribe
    add_word_timestamps(
  File whisper/timing.py", line 323, in add_word_timestamps
    alignment = find_alignment(model, tokenizer, text_tokens, mel, num_frames, **kwargs)
  File whisper/timing.py", line 241, in find_alignment
    start_times = jump_times[word_boundaries[:-1]]
IndexError: arrays used as indices must be of integer (or boolean) type
 ````

### DIFF
--- a/whisper/timing.py
+++ b/whisper/timing.py
@@ -214,6 +214,8 @@ def find_alignment(
     text_indices, time_indices = dtw(-matrix)
 
     words, word_tokens = tokenizer.split_to_word_tokens(text_tokens + [tokenizer.eot])
+    if len(word_tokens) <= 1:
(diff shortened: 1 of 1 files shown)


doublex:
The problem is this line: https://github.com/openai/whisper/blob/main/whisper/timing.py#L217
It returns a array of float (instead int):

````python
np.pad(np.cumsum([]), (1, 0)) 
 -->
[0.] instead [0]
````

PR #1180, opened 2023-04-01
added cantonese to the language list

Keith-Hon:


jongwook:
This unfortunately will not work, as the model is not trained with a separate label for Cantonese. You can use `zh` for Cantonese, and it may work better with some prompting. See 

- #25 
- #366 

for more details, and

- #277 

may also be relevant.

PR #1225, opened 2023-04-12
Add new `job_details.model` key to transcribe return dict

ururk:
Per discussion https://github.com/openai/whisper/discussions/1112

I've added a new key to the return dictionary from the transcribe def:

```JSON
{
    "job_details": {
        "model": "base"
    },
    "text": " And so my fellow Americans, ask not what your country can do for you, ask what you can do for your country.",
    "segments": [{"id": 0, "seek": 0, "start": 0.0, "end": 11.0, "text": " And so my fellow Americans, ask not what your country can do for you, ask what you can do for your country.", "tokens": [50364, 400, 370, 452, 7177, 6280, 11, 1029, 406, 437, 428, 1941, 393, 360, 337, 291, 11, 1029, 437, 291, 393, 360, 337, 428, 1941, 13, 50914], "temperature": 0.0, "avg_logprob": -0.2968965598515102, "compression_ratio": 1.3544303797468353, "no_speech_prob": 0.04440217837691307}], "language": "en"
}
```

Besides `model`, there may be other parameters passed to whisper that could be useful, hence the nested key instead of a top-level property.

### DIFF
--- a/whisper/transcribe.py
+++ b/whisper/transcribe.py
@@ -37,6 +37,7 @@ def transcribe(
     model: "Whisper",
     audio: Union[str, np.ndarray, torch.Tensor],
     *,
+    model_name: str = "small",
     verbose: Optional[bool] = None,
     temperature: Union[float, Tuple[float, ...]] = (0.0, 0.2, 0.4, 0.6, 0.8, 1.0),
     compression_ratio_threshold: Optional[float] = 2.4,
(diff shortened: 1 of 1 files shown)


ururk:
Would you prefer the model be part of the main result dict instead of a in a sub dictionary?

Alfimay19 (approved):

PR #1239, opened 2023-04-16
Formatted Warnings Output

brett-b112:
**Improved Readability of Output From Transcribe**

* Removed the self print of warnings 
   * warnings.warn(string) removed from output
   * Retains useful file pathing associated with issue


jongwook:
I think it'd be better to avoid silently altering a global namespace like `warnings.formatwarning` from within the function. It can be done in the code that calls Whisper.

Was there a specific warning message that was particularly annoying? We can also consider switching to `logging.warning` which is slightly more configurable.

PR #1252, opened 2023-04-17
Add output in TextGrid file support

cmlong02:
- Add WriteTextGrid function in utils.py
- Add "textgrid" option to "output_format" argument

.TextGrid is a common file type used by Praat and other voice analysis softwares. This makes it a perfect fit for Whisper exports, useful to all people who wish to further modify files after running Whisper on them.


jongwook:
Hi, I'm reluctant to add support for a format that I'm not very familiar with. Can you work with a separate script that converts the JSON output into the textgrid format?

ryanheise:
Some ways to convert:

https://pypi.org/project/textgrid-convert/
https://github.com/rctatman/SrtToTextgrid

Maybe you can also make a feature request to Praat to recognise SRT as an import format.

PR #1257, opened 2023-04-19
Teach whisper to look for models in /usr/ and /var/ too.

petterreinholdtsen:
This allow the model to be provided in OS packages and allow users to share model files instead of downloading individual copies to each users home directory.

Existing models in the users home directory is preferred, and files in /var/ are preferred over files in /usr/.  If no model is found in the global paths, the model is downloaded into ~/.cache/whisper/ as before.

### DIFF
--- a/whisper/__init__.py
+++ b/whisper/__init__.py
@@ -46,7 +46,6 @@ _ALIGNMENT_HEADS = {
 
 
 def _download(url: str, root: str, in_memory: bool) -> Union[bytes, str]:
-    os.makedirs(root, exist_ok=True)
 
     expected_sha256 = url.split("/")[-2]
     download_target = os.path.join(root, os.path.basename(url))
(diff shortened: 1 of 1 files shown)


ExtReMLapin:
Just set `WHISPER_CACHE_DIR` env var

petterreinholdtsen:
[ExtReMLapin]
> Just set `WHISPER_CACHE_DIR` env var

Thanks for the idea.  As far as I can tell, it will not provide the
wanted behaviour of preferring shared files while allowing download if
the shared files are missing.

-- 
Happy hacking
Petter Reinholdtsen

jongwook:
I'd like to minimize having platform-specific handling in if possible since I won't be able to test/maintain. Can it be done outside this repo by using `XDG_CACHE_HOME` or some combination of symlinks or shell scripts that can be distributed with your debian package?

petterreinholdtsen:
[Jong Wook Kim]
> I'd like to minimize having platform-specific handling in if possible
> since I won't be able to test/maintain. Can it be done outside this
> repo by using `XDG_CACHE_HOME` or some combination of symlinks or
> shell scripts that can be distributed with your debian package?

I'll see what I can do, but I doubt the same effect can be made without
replicating and trying to stay in sync with the logic inside the
library.

What about making the /usr/ and /var/ locations configuration settings,
to allow the feature to be tested without a UNIX like OS?  Would it help
if I wrote a unit test for this?

-- 
Happy hacking
Petter Reinholdtsen

deinarson:
@petterreinholdtsen thanks for bringing this up, duplicated models have been the bane of my last couple months.

OT but related:  In my case, I have been modifying `transcribe.py` with the code bellow. But I also think that the suggestion of adding tooling for `XDG_CACHE_HOME` is a good idea.

``` python
    if "WHISPER_MODEL_DIR" in os.environ:
      model_dir = os.environ["WHISPER_MODEL_DIR"]
    else:
      model_dir: str = args.pop("model_dir")
```

@ExtReMLapin wrt to `WHISPER_CACHE_DIR` I have greped my entire python lib looking for  it to no avail.  peut-être ai-je mal compris votre suggestion? Did you mean to create a variable for this?

Cheers!
d

PR #1328, opened 2023-05-08
Remove triton dependency on musllinux

pganssle:
`triton` does not currently ship source distributions or musllinux wheels, but `openai-whisper` has a hard pin of `triton==2.0.0` on all `x86_64` Linux targets, making this requirement impossible to satisfy with PyPI alone.

I recommend at least removing the hard pin when not on glibc, though even better would be making `triton` an optional dependency installed with an extra, if possible. Unfortunately, it's [not currently possible](https://discuss.python.org/t/adding-a-default-extra-require-environment/4898) to have an optional depdendency that defaults to "on".

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,8 +12,10 @@ def read_version(fname="whisper/version.py"):
 
 
 requirements = []
+
 if sys.platform.startswith("linux") and platform.machine() == "x86_64":
-    requirements.append("triton==2.0.0")
+    if platform.libc_ver()[0] == "glibc":
+        requirements.append("triton==2.0.0")
 
 setup(
     name="openai-whisper",


jongwook:
Thanks for pointing this out. Would it work if we do `triton>=2,<3`?

pganssle:
> Thanks for pointing this out. Would it work if we do `triton>=2,<3`?

The bigger problem in this case is that `triton` is not shipping source distributions, and they don't have *any* wheels for `musl`. Assuming `whisper` doesnt work with `triton<2`, then `triton>=2` would at least allow building on `musl` if a later version releases `musl` wheels.

Whether or not to do upper version pins is kind of controversial, but I tend to [agree with this posting recommending against it](https://iscinumpy.dev/post/bound-version-constraints/). Ideally you'd use lower bounds reflecting actual dependency requirements and also ship a "known working" lock file, but currently there's no standard way to ship a lock file (the last attempt was [PEP 665](https://peps.python.org/pep-0665/)). You can use [`pip-tools`](https://github.com/jazzband/pip-tools) to generate lock files for a specific platform, and someone can use `pip install whisper -c /path/to/the/lock/file`, but that's not exactly great UI. :sweat:.

AlpinDale:
Now that triton has wheels upstream for musl, I believe this PR should be closed.

HadiSDev:
torch version 2.4 runs on version >=3 So it needs to work there as well

PR #1476, opened 2023-06-29
fix bug with getting dictionary items in utils

bayesfactor:
utils tries to get dictionary items that may not exist, improperly using the `Optional` keyword (not in arguments). Fixes 
[whisper/whisper/utils.py](https://github.com/openai/whisper/blob/248b6cb124225dd263bb9bd32d060b6517e067f8/whisper/utils.py#L104)

Line 104 in [248b6cb](https://github.com/openai/whisper/commit/248b6cb124225dd263bb9bd32d060b6517e067f8)

 raw_max_line_width: Optional[int] = options["max_line_width"]

### DIFF
--- a/whisper/utils.py
+++ b/whisper/utils.py
@@ -101,8 +101,8 @@ class SubtitlesWriter(ResultWriter):
     decimal_marker: str
 
     def iterate_result(self, result: dict, options: dict):
-        raw_max_line_width = options["max_line_width"] if "max_line_width" in options.keys() else None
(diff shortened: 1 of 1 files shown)


jongwook:
It's intended to be supplied with a dictionary, as in:

https://github.com/openai/whisper/blob/5f957da5ca9ab45a4ac149047e6a8004bc9d25bc/whisper/transcribe.py#L454

although I agree that it's not the most intuitive interface if you're using the `SubtitlesWriter` class directly. I think we can consider refactoring this to be `**kwargs` everywhere except the `iterate_result` function.

PR #382, opened 2022-10-21
Uses MPS (Mac acceleration) by default when available

dwarkeshsp:
Currently, Whisper defaults to using the CPU on MacOS devices despite the fact that PyTorch has introduced Metal Performance Shaders framework for Apple devices in the nightly release ([more info](https://pytorch.org/blog/introducing-accelerated-pytorch-training-on-mac/)).

With my changes to __init__.py, torch checks in MPS is available if torch.device has not been specified. If it is, and CUDA is not available, then Whisper defaults to MPS.

This way, Mac users can experience speedups from their GPU by default.

PR #1481, opened 2023-06-29
Add GPU Support for Apple Silicon Macs

NripeshN:


jongwook:
Hi, this is duplicate of #382, where we have various reports that it either fails to run some operations or performance that's not very promising. Looks like it's doing better on pytorch 2.x; I'll plan to merge #382 after confirming that MPS is works reliably and fast enough.

NripeshN:
Hi @jongwook, this pull request may pass the failing operations. PyTorch should be installed with MPS compatibility for torch.backends.mps.is_available() to work. Otherwise, it would throw out an error.

PR #1483, opened 2023-06-30
Avoid rearranging all caches

wangchou:
Since kv_caches from cross attention block is the same for each beam, we can avoid rearranging or calculating it for multiple times. 
I saw about 20% speed up of large model with beam_size = 5 on cpu backend.

### DIFF
--- a/whisper/decoding.py
+++ b/whisper/decoding.py
@@ -145,6 +145,7 @@ class PyTorchInference(Inference):
         self.initial_token_length = initial_token_length
         self.kv_cache = {}
         self.hooks = []
+        self.rearrange_list = [block.attn.key for block in self.model.decoder.blocks] + [block.attn.value for block in self.model.decoder.blocks]
 
     def logits(self, tokens: Tensor, audio_features: Tensor) -> Tensor:
         if not self.kv_cache:
@@ -164,9 +165,13 @@ class PyTorchInference(Inference):
         self.hooks = []
 
     def rearrange_kv_cache(self, source_indices):
-        for module, tensor in self.kv_cache.items():
+        is_same_order = source_indices == list(range(len(source_indices)))
+        if is_same_order:
+            return
+
+        for module in self.rearrange_list:
             # update the key/value cache to contain the selected sequences
-            self.kv_cache[module] = tensor[source_indices].detach()
+            self.kv_cache[module] = self.kv_cache[module][source_indices].detach()
 
 
 class SequenceRanker:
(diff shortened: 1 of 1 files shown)


whisper/decoding.py
```diff
@@ -721,8 +725,7 @@ def run(self, mel: Tensor) -> List[DecodingResult]:
                 )
             ]
 
-        # repeat the audio & text tensors by the group size, for beam search or best-of-n sampling
-        audio_features = audio_features.repeat_interleave(self.n_group, dim=0)
```
yuekaizhang, line 725: @wangchou  I was wondering if you remove the repeat of audio_features, where you repeat the kv_cache for cross attention? Otherwise, during cross_attention, q@k seems with mismatch dims since tokens are repeated according the beam_size.

yuekaizhang: @jongwook, would you mind checking this please? Thanks.

wangchou: @yuekaizhang @ operator (tf.matmul) supports broadcasting since [v1.12.1](https://github.com/tensorflow/tensorflow/commit/47ab68d265a96b6e7be06afd1b4b47e0114c0ee9).

yuekaizhang:
```
import torch
q = torch.ones(70, 4, 16, 4)
k = torch.ones(7, 4, 4, 16)
k2 = torch.ones(70, 4, 4, 16)

context2 = q @ k2
print(context2.shape)

context1 = q @ k
print(context1.shape)
```
I ran with torch==2.0.1
```
RuntimeError: The size of tensor a (70) must match the size of tensor b (7) at non-singleton dimension 0
```

wangchou: @yuekaizhang I ran whisper with beam_size=5. It works.
```
python -m whisper ../samples/thatBand2ch_short.wav --language ja --model small --beam_size=5
```

After adding print in qkv_attention() like
```python
        qk = q @ k
        print("q.shape=",q.shape,", k.shape=", k.shape)
```

it outputs
```
...
q.shape= torch.Size([5, 12, 1, 64]) , k.shape= torch.Size([5, 12, 64, 6])
q.shape= torch.Size([5, 12, 1, 64]) , k.shape= torch.Size([1, 12, 64, 1500])
```

What arguments did you use to get k like (7, 4, 4, 16)? Where is that **7** comes from?

yuekaizhang: @wangchou  Did you try inference with batch_size > 1? I met this issue when I tried with both batch_size, beam_size > 1.

wangchou: @yuekaizhang I don't even know ```batch_size``` option. And I cannot find it with ```whisper --help```. Sorry.

PR #1501, opened 2023-07-05
added parser argument to support MPS by default

NripeshN:
Please merge this PR once merging https://github.com/openai/whisper/pull/382

### DIFF
--- a/whisper/transcribe.py
+++ b/whisper/transcribe.py
@@ -378,12 +378,20 @@ def transcribe(
 def cli():
     from . import available_models
 
+    if torch.cuda.is_available():
+        default_device = "cuda"
+    elif hasattr(torch.backends, 'mps'):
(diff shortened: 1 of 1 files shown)


whisper/transcribe.py
```diff
@@ -378,12 +378,20 @@ def new_segment(
 def cli():
     from . import available_models
 
+    if torch.cuda.is_available():
+        default_device = "cuda"
+    elif hasattr(torch.backends, 'mps'):
+        if torch.backends.mps.is_available():
+            default_device = "mps"
```
rickwierenga, line 385: I think this should be

```
elif hasattr(torch.backends, 'mps') and torch.backends.mps.is_available():
    default_device = "mps"
```

rickwierenga:
LGTM

PR #1511, opened 2023-07-09
Use thread-safe loading of mel matrix IN audio.py

mzamini92:
The `mel_filters` function is using a `np.load` function to load a pre-computed mel filterbank matrix. This function is not thread-safe, which means that if it is called from multiple threads at the same time, it may corrupt the data.

To fix this, you can use the `torch.load` function instead. This function is thread-safe, so it will not corrupt the data if it is called from multiple threads at the same time.


jstoone:
Hi @mzamini92 this is great! I would suggest that you update the PR title, so that it reflects what the overall change strives to address.

Something like:
> Use thread-safe loading of mel matrix

mzamini92:
> Hi @mzamini92 this is great! I would suggest that you update the PR title, so that it reflects what the overall change strives to address.
> 
> Something like:
> 
> > Use thread-safe loading of mel matrix

Thanks for your comment. Done.

jongwook:
Hi, thanks for pointing this out. Is there a documentation stating that `np.load` is not thread-safe? I think using `torch.load()` is a better idea in either case, but the docstring and the `.npz` file should be updated in conjunction.

mzamini92:
> Hi, thanks for pointing this out. Is there a documentation stating that `np.load` is not thread-safe? I think using `torch.load()` is a better idea in either case, but the docstring and the `.npz` file should be updated in conjunction.

Hi. yes 
https://numpy.org/devdocs/reference/generated/numpy.load.html
In the warning it says: Loading files that contain object arrays uses the pickle module, which is not secure against erroneous or maliciously constructed data. Consider passing allow_pickle=False to load data that is known not to contain object arrays for the safer handling of untrusted sources.

PR #1605, opened 2023-08-16
added support for bfloat16 datatype to be used for Intel Xeon processors

devpramod-intel:


devpramod-intel:
Hi @jongwook 
May I get an estimate on the review for this PR?
or if there's a process to be followed? Thank you

jongwook:
Hi, I think it wouldn't hurt, but I don't have access to Xeon hardware to test this. For other platforms, it wouldn't make much sense to compute in BF16 because the model weights are distributed in fp16, and casting to bf16 would just lose some precision, although it's probably okay.

Do you have a performance comparison between fp32 and bf16 inference speeds on Sapphire Rapids?

dbalabka:
@devpramod-intel , it looks that PR is missing proper handling of DecodeOptions in `transcribe()` method:
https://github.com/openai/whisper/blob/main/whisper/transcribe.py#L109-L118

PR #1729, opened 2023-10-22
Add new option to generate subtitles by a specific number of words

amolinasalazar:
Added a new word option called **--max_words_count** that will generate subtitles setting a maximum limit of words per segment. This could sound similar to **--max_line_width** option, but the results are more pleasent for readers IMHO. Here a couple of comparisons using .SRT files:

max_word_count
Notice that **--max_words_count** works as an upper bound of words, but still it will respect the segments in the way that end of sentences can have less words if the remaining number of words in a segment is lower than the max_words_count value.
i.e. _Segment = [word1, word2, word3, word4, word5] and max_words_count = 3
=>Result = [word1, word2, word3] and [word4, word5]_
This is not the behaviour we can see using **--max_line_width** that can leave bigger gaps of time when joining end and beginning of segments:
WidthvsWords

Subtitles generated with **--max_words_count** look similar of what we can see in Shorts, Reels and other short duration videos.

This is my first contribution, so feel free of changing/comment/improve anything.

### Additional notes

- The use  of **--max_line_width** will disable the effects of **--max_words_count**.
- Manually tested using Python and cli and checked results in .srt and .vtt files (.txt. and .tsv files won't be affected).

### DIFF
--- a/whisper/transcribe.py
+++ b/whisper/transcribe.py
@@ -452,7 +452,12 @@ def cli():
     model = load_model(model_name, device=device, download_root=model_dir)
 
     writer = get_writer(output_format, output_dir)
-    word_options = ["highlight_words", "max_line_count", "max_line_width", "max_words_per_line"]
+    word_options = [
+        "highlight_words",
+        "max_line_count",
+        "max_line_width",
+        "max_words_per_line",
+    ]
     if not args["word_timestamps"]:
         for option in word_options:
             if args[option]:
--- a/whisper/transcribe.py
+++ b/whisper/transcribe.py
@@ -412,6 +412,7 @@ def cli():
     parser.add_argument("--highlight_words", type=str2bool, default=False, help="(requires --word_timestamps True) underline each word as it is spoken in srt and vtt")
     parser.add_argument("--max_line_width", type=optional_int, default=None, help="(requires --word_timestamps True) the maximum number of characters in a line before breaking the line")
     parser.add_argument("--max_line_count", type=optional_int, default=None, help="(requires --word_timestamps True) the maximum number of lines in a segment")
+    parser.add_argument("--max_words_count", type=optional_int, default=None, help="(requires --word_timestamps True) (no effect with --max_line_width) the maximum number of words in a segment")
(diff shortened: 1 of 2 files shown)


FurkanGozukara:
amazing

khaledbkheet:
from pydub import AudioSegment

song = AudioSegment.from_mp3("good_morning.mp3")

# PyDub handles time in milliseconds
ten_minutes = 10 * 60 * 1000

first_10_minutes = song[:ten_minutes]

first_10_minutes.export("good_morning_10.mp3", format="mp3")

FurkanGozukara:
@amolinasalazar which word count do you suggest for youtube?

amolinasalazar:
> @amolinasalazar which word count do you suggest for youtube?

Actually I think that's a really personal choice and it can depend on several things.

In the end, the main factor why Reels or Shorts normally have just a couple of words on the screen at a moment is because of the aspect ratio of the videos. Having long subtitle lines for videos watched in mobile phones with a vertical orientation will fill the whole screen with words.

There are other factors like the font size, the speed of the speech or even the complexity of the context. Less words normally create dynamic and impactful videos, ideal for simple and strong messages, but it can be stressful if those have a long duration. For example, I won't set 1-3 words at a time if you are explaining a hard topic as it could be stressful to understand.

So in my opinion, you need to find a comfortable number on your own, but something between 3-6 words can be pleasant in general.

demoskalifi:
can I use this using the openai whisper API? if so, how?

Francoyy:
The command name is `--max_words_per_line` and not `--max_words_count` (https://github.com/openai/whisper/pull/1729/files#diff-f6accbbb4ebcd3dd6815bf012490d9ba37eb89a65f2124adc95c2a39bc6941b7R422)
An example of command would be
`whisper file.mp4 --language English --model large-v3 --output_format srt --max_words_per_line 6`

amolinasalazar:
> The command name is `--max_words_per_line` and not `--max_words_count` (https://github.com/openai/whisper/pull/1729/files#diff-f6accbbb4ebcd3dd6815bf012490d9ba37eb89a65f2124adc95c2a39bc6941b7R422) An example of command would be `whisper file.mp4 --language English --model large-v3 --output_format srt --word_timestamps True --max_words_per_line 6`

True, jonwook renamed this command before merging the PR. I'll update the first comment so there are no confussions.

PR #1802, opened 2023-11-13
Relax triton requirements for compatibility with pytorch 2.1 and newer

eindenbom:
PyPI package of PyTorch 2.1 uses triton==2.1.0, which makes it incompatible with whisper due to overly strict requirement triton==2.0.0.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,7 @@ def read_version(fname="whisper/version.py"):
 
 requirements = []
 if sys.platform.startswith("linux") and platform.machine() == "x86_64":
-    requirements.append("triton==2.0.0")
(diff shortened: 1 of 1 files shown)


eindenbom:
@jongwook, what is the schedule for next PyPI release?

setup.py
```diff
@@ -13,7 +13,7 @@ def read_version(fname="whisper/version.py"):
 
 requirements = []
 if sys.platform.startswith("linux") and platform.machine() == "x86_64":
-    requirements.append("triton==2.0.0")
```
miracle1015, line 16: -    requirement.append("triton>=2.0.0,<3")

PR #1857, opened 2023-11-30
Replace regular expressions

eukub:
to reduce redundancy and improve performance


ExtReMLapin:
For performance, use re.compile

eukub:
> For performance, use re.compile

also it's a way to process not only ASCII chars but unicode too

PR #1866, opened 2023-12-02
detect if Mac's Metal Performance Shaders are available

rgbkrk:
Default to [Torch's `mps` backend](https://pytorch.org/docs/stable/notes/mps.html) if available when `cuda` is not.


rgbkrk:
Closing as https://github.com/openai/whisper/pull/382 already exists.

PR #1959, opened 2024-01-15
add repeat audio back

yuekaizhang:
Related to #1483.

I have encountered issues when decoding with beam_size > 1. 

Here, https://github.com/openai/whisper/blob/main/whisper/model.py#L102, during cross_attention, I think we must keep the first two dims of query and key tensors same. However, I didn't find anywhere we repeat the audio or key, value tensors with beam_size.

### DIFF
--- a/whisper/decoding.py
+++ b/whisper/decoding.py
@@ -730,7 +730,8 @@ class DecodingTask:
                 )
             ]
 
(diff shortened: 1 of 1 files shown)


Theghost256 (approved):

yuekaizhang:
@Theghost256  Thanks, would you mind approving the CI pipelien please?

PR #1979, opened 2024-01-25
Fix model.transcribe

jbetker:
Some of the inference functions now return lists of dicts where they previously returned just dicts, causing them to fail when transcribe is called. This makes a trivial fix of just picking the first list option when that occurs.


jbetker:
The MEL input was batched.

PR #2004, opened 2024-02-06
Add language_bias parameter to detect_language

jbaudanza:
We are doing a lot of multi-lingual transcriptions, especially with language learners. We have found that Whisper will often incorrectly identify the language when a language learner is speaking their target language.

For example, a native Korean speaker will speak English, but with a strong Korean accent, and Whisper will identify the language as Korean, not English.

We've found that adding a language bias of +1.0 to the language learner's target language is enough to nudge whisper in the direction of the user's target language, while still allowing it to properly identify the user's native language, and other languages.

### DIFF
--- a/whisper/decoding.py
+++ b/whisper/decoding.py
@@ -17,7 +17,7 @@ if TYPE_CHECKING:
 
 @torch.no_grad()
 def detect_language(
-    model: "Whisper", mel: Tensor, tokenizer: Tokenizer = None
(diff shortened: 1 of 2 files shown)


jmgb27:
Hi @jbaudanza did you have any workarounds for this one? looks like they haven't fixed it yet.

jbaudanza:
> Hi @jbaudanza did you have any workarounds for this one? looks like they haven't fixed it yet.

This PR is the workaround. 

Alternatively, you could try using a different model to do the language detection part. Maybe try this: https://huggingface.co/speechbrain/lang-id-voxlingua107-ecapa I haven't compared the results to Whisper though.

PR #2005, opened 2024-02-07
Prevent infinite loop for out-of-bound timestamps in clip_timestamps

jbaudanza:
I'm using the output of `get_speech_timestamps` in Silero VAD to produce values for `clip_timestamps`.

The VAD produces an end timestamp that is one-sample beyond the last found voice sample. I was converting these timestamps into seconds and passing them directly into `clip_timestamps`.

The problem is sometimes the end timestamp would be one sample past the end of the file. In this case, whisper would go into an infinite loop, trying to run inference on segment_size of 0.

### DIFF
--- a/whisper/transcribe.py
+++ b/whisper/transcribe.py
@@ -261,6 +261,8 @@ def transcribe(
         #     while seek < seek_clip_end
         while clip_idx < len(seek_clips):
             seek_clip_start, seek_clip_end = seek_clips[clip_idx]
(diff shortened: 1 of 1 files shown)


louisbrulenaudet (approved):

ryanheise:
It looks like `whisper-test` has gotten stuck for some reason.

Theghost256 (approved):

PR #2097, opened 2024-03-20
Hiding the ffmpeg popup that appears momentarily

Menonro:
On Windows there is a problem due to which FFMPEG running in the background becomes small for a second during startup.

Context: "Whisper" is used in conjunction with FastAPI and uvicorn. The service is started using pm2.

The problem does not exist if you run it manually from a regular console using "python main.py"

Demonstration of the problem before and after
https://youtu.be/IsEavmzQOv8

### DIFF
--- a/whisper/audio.py
+++ b/whisper/audio.py
@@ -1,6 +1,6 @@
 import os
 from functools import lru_cache
-from subprocess import CalledProcessError, run, CREATE_NO_WINDOW
+from subprocess import CalledProcessError, run
 from typing import Optional, Union
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


glangford (commented):
CREATE_NO_WINDOW is only valid on Windows platforms. The import fails for other OS's.

https://stackoverflow.com/questions/25122823/hide-console-with-subprocess-popen

Menonro:
> CREATE_NO_WINDOW действителен только на платформах Windows. Импорт не удался для других ОС.
> 
> https://stackoverflow.com/questions/25122823/hide-console-with-subprocess-popen

I made a commit based on your comment.

Menonro:
> This line of the proposed change
> 
> ```python
> from subprocess import CalledProcessError, run, CREATE_NO_WINDOW
> ```
> 
> will fail with `ImportError: cannot import name 'CREATE_NO_WINDOW' from 'subprocess'` on non-Windows platforms.

Thank you for your patience, I will install myself a virtual machine for testing. I'm testing it on my Windows. Made a correction so that import occurs only on Windows systems.

Theghost256 (approved):

PR #2189, opened 2024-05-26
Add probability for each token

khalilfiremind:
Thanks to @SinanAkkoyun https://github.com/SinanAkkoyun/openai-whisper/tree/main, I modified his code a bit, so it can work with `large-v3`

The `stt` method will return the text tokens with the proba for each token. This can be helpful in application like pronunciation checking.

How to use?

```
import whisper
from whisper import stt, load_audio

PATH_TO_AUDIO = ""
model = whisper.load_model("large-v3")
audio = whisper.load_audio( PATH_TO_AUDIO)

# this will return list of lists, first item in each list will be the token and the second item will be the probability
result = stt(model, audio)
```

### DIFF
--- a/examples/test_prob.py
+++ b/examples/test_prob.py
@@ -2,13 +2,28 @@ import sys
 from pathlib import Path
 sys.path.insert(0, str(Path(__file__).resolve().parents[1]))
 
+import torch
 import whisper
 import argparse
 import colorsys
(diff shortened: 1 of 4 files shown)


SinanAkkoyun:
Thank you very much 😄🤗

Theghost256 (approved):

PR #2256, opened 2024-07-02
Support longer audio files reducing memory usage with chunking

ggarber:
## Description
The current implementation of Whisper loads the entire audio file into memory during transcription. This approach demands a substantial amount of memory, making it highly likely to run out of memory, especially for very long audio files (many hours).

This proposed change modifies the processing mechanism so that the audio file, when processed by ffmpeg, is loaded in chunks with a maximum duration of two hours. These chunks are processed sequentially, significantly reducing memory usage.

## Test Plan
Unit Tests: Verified with existing unit tests to ensure no regression.
Long Audio Files: Tested with audio files up to 16 hours in duration to confirm stability and efficiency.


kentslaney:
I have a [real-time transcription repo](https://github.com/kentslaney/realtime-whisper) that would also solve this problem by limiting the number of bytes stored in memory

PR #2265, opened 2024-07-15
Support for Buffered Audio Data

kentslaney:
This pull request adds the CLI flag `--buffered`, which allows the audio data to be loaded on-demand through an I/O buffer instead of all at once at the beginning. This provides an alternative resolution to the problem addressed in #2256. The default memory limit is dictated [here](https://github.com/kentslaney/openai-whisper/blob/c09790488bfa5a2fbf9d7decbd8d5747df4f5b27/whisper/buffer.py#L36), but shouldn't matter much since it's not a bottleneck. 

Support for buffered input also paves the way towards real-time tasks (#2): the changes were directly adapted from [kentslaney/realtime-whisper](https://github.com/kentslaney/realtime-whisper/), which already supports live PCM data. In order to keep the PR from being any larger than it already is, changes were limited to the minimum motivating feature. I'm making a PR instead of keeping it as a wrapper because it reuses/refactors large chunks of `transcribe`, which the current scoping makes inaccessible for patching.

The buffered version will perform almost exactly the same set of operations, with a small (vanishing) difference in the preprocessing output. The log mel spectogram is clamped below to have a maximum range of 8. The original `transcribe` method computes the upper bound from the entire file's spectogram, while the buffered version clamps the values relative to the maximum in the chunks processed so far. I have done [some regression testing](https://github.com/kentslaney/realtime-whisper/blob/be37824199610a7b2e946f307b938d441970143c/test.py#L142-L168) to confirm that the refactored (non-buffered) solution will make the same calls as the existing one.

I realize this is a rather large PR. Please let me know if these changes are outside the scope of the repo or if there's anything I can do to make it easier to review.

### DIFF
--- a/whisper/batching.py
+++ b/whisper/batching.py
@@ -1,14 +1,18 @@
-import torch
-import numpy as np
-from collections.abc import Callable, AsyncIterable, AsyncIterator, Awaitable
+from collections.abc import AsyncIterable, AsyncIterator, Awaitable, Callable
 from typing import Generic, TypeVar, Union
 
+import numpy as np
+import torch
+
 A = TypeVar("A", bound=Union[np.ndarray, torch.Tensor])
+
 
 class ArrayWrapper(Generic[A]):
     pass
 
+
 ArrayTypes = Union[A, ArrayWrapper[A]]
+
 
 class LoopbackIterator(Generic[A]):
     async def iter(self):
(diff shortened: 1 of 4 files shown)


gabriel370pp:
Uploading Screenrecorder-2024-07-22-02-17-04-378.mp4…

gabriel370pp:
Vv

Theghost256 (approved):

PR #2298, opened 2024-08-08
Pin numpy to 1.26.4

agardnerIT:
whisper is not compatible with numpy 2.x so pin to highest 1.x version

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numba
-numpy
+numpy==1.26.4
 torch
 tqdm
 more-itertools


x86Gr:
Can it be made compatible with numpy 2.x ?

agardnerIT:
> Can it be made compatible with numpy 2.x ?

That's beyond my knowledge. This PR fixes everyone's Quickstart experience.

cyberlightdev:
+1

In the meantime, `pip install --upgrade numpy==1.26.4` does the job as well, but this quick fix would streamline things for non-Pytrhon devs like myself.

Thanks for the clue about the cryptic error I received running with Numpy 2.0.1.  I didn't mean to learn Python today, but here we are.

Theghost256 (approved):

PR #2301, opened 2024-08-13
Fix/torch load weights only warning

Ultr4Dev:
This pull request updates the load_model function of the Whisper ASR model, enhancing documentation and incorporating a weights_only parameter for more flexible model weight handling. The update addresses a FutureWarning regarding the use of torch.load with weights_only=False, recommending a safer loading method to mitigate security risks associated with loading untrusted model data. This change is designed to improve both the functionality and security of the model loading process.

Relevant info [Untrusted models](https://github.com/pytorch/pytorch/blob/main/SECURITY.md#untrusted-models)


glangford:
Just to add to the documentation of this issue, the default for `torch.load()` is `weights_only=False` back to at least 2.0. So the default behaviour is that torch.load

> ...  uses (the) `pickle` module implicitly, which is known to be insecure. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling

(from https://pytorch.org/docs/2.4/generated/torch.load.html#torch-load)

Related
https://github.com/pytorch/pytorch/issues/52181

Ultr4Dev:
> Just to add to the documentation of this issue, the default for `torch.load()` is `weights_only=False` back to at least 2.0. So the default behaviour is that torch.load
> 
> > ...  uses (the) `pickle` module implicitly, which is known to be insecure. It is possible to construct malicious pickle data which will execute arbitrary code during unpickling
> 
> (from https://pytorch.org/docs/2.4/generated/torch.load.html#torch-load)
> 
> Related [pytorch/pytorch#52181](https://github.com/pytorch/pytorch/issues/52181)
> 
> HuggingFace on Pickle https://huggingface.co/docs/hub/en/security-pickle

Thank you for the addition!

Theghost256 (approved):

vblazenka:
Hey @Ultr4Dev thanks for creating this PR. Do you plan to move it forward? 🙏

jongwook:
thanks! #2451 fixed this

PR #2307, opened 2024-08-19
Relax triton requirements for compatibility with pytorch 2.4 and newer

xingjianan:
Similar to https://github.com/openai/whisper/pull/1802, but now when pytorch upgrades to 2.4, it requires triton==3.0.0. I am not sure if it makes sense to remove the upper bound version constraints


Kvadratni:
@jongwook Is there any chance this can be merged?

PR #2359, opened 2024-09-30
using sdpa if available

jongwook:


JonahMMay:
I have the same issue as @formater where is_causal is receiving a Tensor instead of a bool.

affinanti333 (commented):
Final

PR #2361, opened 2024-09-30
large-v3-turbo model

jongwook:


FurkanGozukara:
@jongwook awesome

is this better or just faster?

jongwook:
just faster. will share more details later today

PR #2379, opened 2024-10-07
Update README.md

BotMaster3000:
Default now uses Turbo instead of Small


BotMaster3000:
Noticed it uses Turbo instead of Small by now, had me confused so I changed the Readme to reflect that.

Kryslynn93 (commented):
Approved

PR #2390, opened 2024-10-15
Improvements implemented in the audio processing module

mrfelpa:
- The relative import causing errors when the script was executed directly, so I changed from .utils import exact_div to from utils import exact_div. I also implemented a function ***(get_hann_window)*** to avoid repeated Hann window calculations.

### DIFF
--- a/whisper/audio.py
+++ b/whisper/audio.py
@@ -7,41 +7,39 @@ import numpy as np
 import torch
 import torch.nn.functional as F
 
-from .utils import exact_div
+# Importação corrigida
+from utils import exact_div
 
-# hard-coded audio hyperparameters
-SAMPLE_RATE = 16000
-N_FFT = 400
-HOP_LENGTH = 160
-CHUNK_LENGTH = 30
-N_SAMPLES = CHUNK_LENGTH * SAMPLE_RATE  # 480000 samples in a 30-second chunk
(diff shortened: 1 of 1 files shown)


whisper/audio.py
```diff
@@ -53,105 +51,175 @@ def load_audio(file: str, sr: int = SAMPLE_RATE):
...
+   
+   magnitudes = stft.abs() ** 2
+
+   # Calculate Mel spectrogram and apply logarithmic scaling.
+   filters = mel_filters(audio_tensor.device, n_mels)
+   mel_spec = filters @ magnitudes
+
+   log_spec = torch.clamp(mel_spec, min=1e-10).log10()
+   
+   log_spec_normalized = (log_spec + 4.0) / 4.0
```
Nisarg236, line 223: maybe combine both lines?
`log_spec_normalized = (torch.clamp(mel_spec, min=1e-10).log10() + 4.0) / 4.0
`

Nisarg236, line 223: maybe combine both?
log_spec_normalized = (torch.clamp(mel_spec, min=1e-10).log10() + 4.0) / 4.0

mrfelpa: @Nisarg236 I did it separately to try to keep the code clearer, but yes, it is possible to combine the two.

PR #2421, opened 2024-11-01
Add compression_ratio_hallucination_threshold

Ko4ka:
Add compression_ratio_hallucination_threshold to Discard High Compression Ratio Segments in transcribe()

Here is the detailed description

https://github.com/openai/whisper/discussions/2420


whisper/transcribe.py
```diff
@@ -42,11 +42,11 @@ def transcribe(
     verbose: Optional[bool] = None,
     temperature: Union[float, Tuple[float, ...]] = (0.0, 0.2, 0.4, 0.6, 0.8, 1.0),
     compression_ratio_threshold: Optional[float] = 2.4,
+    compression_ratio_hallucination_threshold: Optional[float] = 3,
     logprob_threshold: Optional[float] = -1.0,
     no_speech_threshold: Optional[float] = 0.6,
     condition_on_previous_text: bool = True,
     initial_prompt: Optional[str] = None,
-    carry_initial_prompt: bool = False,
```
apetresc, line 49: I don't think you meant to remove this?

whisper/transcribe.py
```diff
@@ -205,7 +208,7 @@ def decode_with_fallback(segment: torch.Tensor) -> DecodingResult:
                 compression_ratio_threshold is not None
                 and decode_result.compression_ratio > compression_ratio_threshold
             ):
-                needs_fallback = True  # too repetitive
+                needs_fallback = True  # too repetitive <-- We can inprove it...
```
apetresc, line 211: "inprove" should be "improve" (although to be honest I'm not sure what this entire comment is trying to convey)

whisper/transcribe.py
```diff
@@ -216,6 +219,13 @@ def decode_with_fallback(segment: torch.Tensor) -> DecodingResult:
                 and decode_result.no_speech_prob > no_speech_threshold
             ):
                 needs_fallback = False  # silence
+            if (
+            compression_ratio_hallucination_threshold is not None
```
apetresc, line 223: The condition should be indented one more level to match the style of the multi-line if conditions around it)

apetresc (commented):
Minor comments

whisper/transcribe.py
Ko4ka: True, fixed in the latest commit

whisper/transcribe.py
Ko4ka: Forgot this comment fixed in the latest commit

whisper/transcribe.py
Ko4ka: Fair enough, fixed

Ko4ka:
@apetresc thx
I've updated the code

nicholasgcotton:
This would be of great value for my org, as our audio often has extensive silent periods when people leave and re-enter the interview suites to go off the record. This is a great solution.

montvid:
Does not solve the problem completely in my tests. Only VAD fixed the hallucinations. Interesting that VLLM server Whisper implementation feature does expose the compression_ratio in its JSON.

PR #2451, opened 2024-11-20
Fix: Update torch.load to use weights_only=True to prevent security w…

yaslack:
#### **Changes Made:**
In this pull request, I updated the `torch.load` function call to use `weights_only=True` in order to prevent a security warning related to the unpickling process.

#### **Code Change:**

```diff
- checkpoint = torch.load(fp, map_location=device)
+ checkpoint = torch.load(fp, map_location=device, weights_only=True)

### CODE REVIEW

**Comment by bitplane** [comment]
For anyone else with this issue, I've been doing this for now:

```python
import functools
whisper.torch.load = functools.partial(whisper.torch.load, weights_only=True)
```

hamirmahal (commented):
@jongwook this seems like a useful change that would reduce a lot of noise for developers.

Are you open to it?

bitplane:
> For anyone else with this issue, I've been doing this for now:
> 
> ```python
> import functools
> whisper.torch.load = functools.partial(whisper.torch.load, weights_only=True)
> ```

This was happening in other things too, so I ended up making a module for it. Trivial, but I found it useful so thought I'd share:

https://github.com/bitplane/torch_weightsonly

oscardssmith:
any updates here? this seems like an improvement that should be merged

jongwook:
#505 was for the same issue but i didn't want to break compatibility with torch==1.10.1 just for this. The commit above should make it work for all version.

PR #2452, opened 2024-11-21
Add support for leading zero in transcription when spoken by user

chihangc:
This pull request addresses an issue where leading zeros in numbers were not preserved in the transcription when explicitly spoken by the user. e.g. AB01234 was normalized as AB 1234. The changes ensure that the transcription includes leading zeros when they are intentionally voiced, by add back the zero after `to_fraction`


AnhMeo:
Nice! What was it that caused the leading zeroes to get removed in the first place?

PR #2580, opened 2025-04-24
Update requirements.txt

Thanmai939:


ExtReMLapin:
Want to read something scary ?

Keep in mind only 50% of india has access to the internet.

PR #2621, opened 2025-07-13
Avoid loading large models into host memory when in_memory is False.

chadf:
Currently the entire model is always loaded into memory to perform hash, and once again if a new model is downloaded. This could have a significant impact on memory usage when large models are used, only to immediately discard.

### DIFF
--- a/whisper/__init__.py
+++ b/whisper/__init__.py
@@ -3,7 +3,7 @@ import io
 import os
 import urllib
 import warnings
-from typing import List, Optional, Union
+from typing import List, Optional, Union, BinaryIO
 
 import torch
 from tqdm import tqdm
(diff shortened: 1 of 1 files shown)


OrangeDoro (commented):
Hi! I'm a grad student working on a research project about using large language models to automate code review. Based on your commit 578c01b83ccc947b86e3551fe7939278c1f3747d and the changes in whisper/__init__.py, my tool generated this comment:
1. In the `_download` function, raise an exception instead of issuing a warning when the SHA256 checksum does not match to prevent the use of potentially corrupted files.
2. Ensure that the file pointer is closed properly regardless of the `in_memory` flag in the `load_model` function to prevent resource leaks.
3. In the `_download` function, validate the `url` and `root` parameters to ensure they are valid before proceeding with the download.
4. Wrap file operations in the `_download` function with exception handling to catch potential `IOError` or `OSError`.
5. In the `load_model` function, check that `checkpoint_file` is a valid file before attempting to open it.
6. Wrap the `torch.load` operation in a try-except block in the `load_model` function to catch potential exceptions and provide a user-friendly error message.
7. Correct the typo in the `RuntimeError` raised when the SHA256 checksum does not match from "does not not match" to "does not match."
8. Provide more context in the error message when raising a `RuntimeError` for a model not found in `_MODELS`, including the name of the model that was not found.
9. Ensure that the `alignment_heads` variable is set appropriately regardless of how the model is loaded, as it is assigned but not used when loading from a file.
10. Ensure compatibility with Python 3.11 or later, or provide a fallback mechanism for earlier versions, as the `hashlib.file_digest` function is defined conditionally.
11. Add logging when a file is being downloaded or when a checksum mismatch occurs to improve observability.
12. Consider adding unit tests for `_file_digest` and `_download` to verify their correctness and handle various edge cases.

As part of my research, I'm trying to understand how useful these comments are in real-world development. If you have a moment, I'd be super grateful if you could quickly reply to these two yes/no questions:


1. Does this comment provide suggestions from a dimension you hadn’t considered?

2. Do you find this comment helpful?

Thanks a lot for your time and feedback! And sorry again if this message is a bother.

iamkallolpratim:
@revux

PR #2730, opened 2026-02-11
fix: raise TypeError instead of returning it

Mr-Neutr0n:
## Bug
`return TypeError(...)` on line 660 of `whisper/decoding.py` silently returns an exception object as the function's return value instead of actually raising the error. This means callers receive a TypeError object instead of getting a proper exception, which would cause confusing downstream failures.

## Fix
Changed `return TypeError(...)` to `raise TypeError(...)` so the error is properly raised when `audio_features` has an incorrect dtype.

### DIFF
--- a/whisper/decoding.py
+++ b/whisper/decoding.py
@@ -657,7 +657,7 @@ class DecodingTask:
         if audio_features.dtype != (
             torch.float16 if self.options.fp16 else torch.float32
         ):
-            return TypeError(
+            raise TypeError(
                 f"audio_features has an incorrect dtype: {audio_features.dtype}"
             )
 


Mr-Neutr0n:
Following up on this one. The change in `whisper/decoding.py` swaps `return TypeError(...)` for `raise TypeError(...)`, a subtle but important fix since returning an exception object instead of raising it means calling code won't actually catch the error. The diff is just +1/-1 and the branch is mergeable. Are you still interested in seeing this land?

Mr-Neutr0n:
Closing this one to tidy up my open pull requests.

It's been open around five months with no review activity, and I'd rather withdraw it than leave it sitting in your queue. Nothing needed from you, and no hard feelings at all.

If the fix is still wanted, this can be reopened, or I'm happy to redo it properly against current main. Apologies for the noise.

PR #2760, opened 2026-04-13
Pin pre-commit hook revisions to immutable commits

mcgrew-oai:
## What's changing
Replace non-commit `rev:` values in `.pre-commit-config.yaml` with the exact commit those refs resolve to today.

## Why
Tags and branch names can be retargeted upstream; pinning to immutable commits keeps the selected code the same over time.


dima55455:
Apk

dima55455:
Apk download ssr from u and keys

PR #2793, opened 2026-06-16
fix: get_start/get_end raise KeyError when segments lack 'words' key

JSap0914:
## Bug

`get_start` and `get_end` in `whisper/utils.py` use `s["words"]` inside a generator passed to `next()`.  When a segment has no word-level timestamps — the common case when `word_timestamps=False` — the generator raises `KeyError: 'words'` instead of exhausting and returning the intended fallback.

```python
>>> from whisper.utils import get_start, get_end
>>> segs = [{'start': 1.0, 'end': 3.0, 'text': 'hello'}]
>>> get_start(segs)   # KeyError: 'words'
>>> get_end(segs)     # KeyError: 'words'
```

## Fix

Replace `s["words"]` with `s.get("words", [])` in both generator expressions. An absent or empty `words` list produces no items, so `next()` falls through to the correct fallback (`segments[0]["start"]` / `segments[-1]["end"]`).

## Verification

```
pytest tests/test_utils.py -v
# 12 passed in 0.06s
```

The new `tests/test_utils.py` covers both the regression and adjacent helpers (`format_timestamp`).

> AI-assisted: this fix was identified and implemented with AI coding assistance.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,104 @@
+"""Tests for pure-Python helpers in whisper/utils.py."""
+import importlib.util
+
+import pytest
+
+
+def _load_utils():
+    spec = importlib.util.spec_from_file_location(
+        "whisper_utils",
+        __file__.replace("tests/test_utils.py", "whisper/utils.py"),
+    )
+    mod = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(mod)
+    return mod
+
+
+utils = _load_utils()
+
+
+# ---------------------------------------------------------------------------
+# format_timestamp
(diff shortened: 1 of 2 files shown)


kiwigitops (commented):
One portability issue in the new test helper: `__file__.replace(tests/test_utils.py, whisper/utils.py)` only matches POSIX-style paths. On Windows, `__file__` will normally contain backslashes, so the replacement does not happen and `spec_from_file_location()` points back at `tests/test_utils.py`.

Using `pathlib`, for example `Path(__file__).resolve().parents[1] / whisper / utils.py`, would make this independent of the path separator.

PR #2795, opened 2026-06-21
fix: don't suppress fallback when output is a repetition loop

tahirkhan05:
# fix: don't suppress fallback when output is a repetition loop

Fixes the root cause of the hallucination/repetition loop bug reported in
discussions #192 and #2442. Also related to PR #1903 (2023) which identified
this issue but was never merged.

## What's happening

`decode_with_fallback` retries decoding at higher temperatures when it detects
bad output. There are two triggers for a retry:

- `compression_ratio > threshold` — output is too repetitive
- `avg_logprob < threshold` — output confidence is too low

There's also a guard that says "if `no_speech_prob` is high and `avg_logprob`
is low, this is silence — don't bother retrying":

```python
if (
    no_speech_threshold is not None
    and decode_result.no_speech_prob > no_speech_threshold
    and logprob_threshold is not None
    and decode_result.avg_logprob < logprob_threshold
):
    needs_fallback = False  # silence
```

The problem: this guard runs **after** `needs_fallback` has already been set
to `True` by the repetition detector, and it unconditionally resets it to
`False`. So when the model is stuck in a loop ("Thank you. Thank you. Thank
you."), the sequence is:

1. `compression_ratio` is very high → `needs_fallback = True`
2. `avg_logprob` is low (garbage tokens) → `needs_fallback = True`  
3. `no_speech_prob` is elevated (model is confused) → silence guard fires → `needs_fallback = False`
4. Retry never happens. Loop continues forever.

## Fix

The silence guard should only fire when the output is not already flagged as a
repetition loop. One extra condition:

```python
if (
    no_speech_threshold is not None
    and decode_result.no_speech_prob > no_speech_threshold
    and logprob_threshold is not None
    and decode_result.avg_logprob < logprob_threshold
    and (
        compression_ratio_threshold is None
        or decode_result.compression_ratio <= compression_ratio_threshold
    )
):
    needs_fallback = False  # silence
```

If `compression_ratio` is above threshold, the output is a repetition loop —
not silence. The retry must happen.

## Testing

Added `tests/test_fixes.py` with 8 unit tests covering the decision logic.
No model weights needed to run them.

```
pytest tests/test_fixes.py -v
```

Closes #192 (partially), related to #2442, references #1903.

### DIFF
--- a/tests/test_fixes.py
+++ b/tests/test_fixes.py
@@ -1,0 +1,162 @@
+import types
+
+import pytest
+
+
+# ---------------------------------------------------------------------------
+# whisper/tokenizer.py -- Javanese language code alias
+# ---------------------------------------------------------------------------
+
+
+def test_jv_alias_in_to_language_code():
+    from whisper.tokenizer import TO_LANGUAGE_CODE
+
+    assert "jv" in TO_LANGUAGE_CODE, (
+        "'jv' (ISO 639-1 code for Javanese) must be in TO_LANGUAGE_CODE so "
+        "users can pass --language jv"
+    )
+    assert TO_LANGUAGE_CODE["jv"] == "jw", (
+        "'jv' must map to 'jw' to match the token the model was trained with"
+    )
+
+
+def test_jw_still_in_languages():
+    # The LANGUAGES key must stay as "jw" -- the model checkpoint uses <|jw|>
+    from whisper.tokenizer import LANGUAGES
+
+    assert "jw" in LANGUAGES
+    assert LANGUAGES["jw"] == "javanese"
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
The regression tests don't actually exercise the production fallback path. `_needs_fallback()` in `tests/test_fixes.py` is a copy of the new condition, so `decode_with_fallback()` could regress and these tests would still pass. The same file also carries unrelated Javanese/progress tests. I'd test the real `decode_with_fallback` path with controlled decode results and keep this PR's test file scoped to the fallback change.

PR #2797, opened 2026-06-21
feat: add progress_callback parameter to transcribe()

tahirkhan05:
# feat: add progress_callback parameter to transcribe()

Adds an optional `progress_callback` parameter to `transcribe()` so callers
can track transcription progress without having to parse `tqdm` output or
redirect stdout.

This has been requested in PRs #2469 and #2625, both of which are open but
have no tests and don't handle the `None` case cleanly.

## API

```python
def transcribe(
    model,
    audio,
    *,
    progress_callback: Optional[Callable[[int, int], None]] = None,
    ...
):
```

The callback receives `(current_frame, total_frames)` on each seek update.
Callers can derive percentage as `current / total`.

Example:

```python
def on_progress(current, total):
    print(f"{current / total:.0%}")

result = whisper.transcribe(model, "audio.mp3", progress_callback=on_progress)
```

The callback is purely optional — passing `None` (the default) leaves all
existing behaviour unchanged.

## Changes

- `whisper/transcribe.py`: add `progress_callback` parameter, call it inside
  the seek loop after `pbar.update()`
- `tests/test_fixes.py`: 3 tests covering callback invocation count, `None`
  safety, and correct fraction values

## Testing

```
pytest tests/test_fixes.py -v
```

Related to #2469, #2625.

### DIFF
--- a/tests/test_fixes.py
+++ b/tests/test_fixes.py
@@ -1,0 +1,162 @@
+import types
+
+import pytest
+
+
+# ---------------------------------------------------------------------------
+# whisper/tokenizer.py -- Javanese language code alias
+# ---------------------------------------------------------------------------
+
+
+def test_jv_alias_in_to_language_code():
+    from whisper.tokenizer import TO_LANGUAGE_CODE
+
+    assert "jv" in TO_LANGUAGE_CODE, (
+        "'jv' (ISO 639-1 code for Javanese) must be in TO_LANGUAGE_CODE so "
+        "users can pass --language jv"
+    )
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
The progress tests never call `transcribe()` or the patched seek loop; they just call a local callback from a test-only `for` loop. They'd stay green if the production `progress_callback(...)` call were removed entirely. Also, this test file includes unrelated Javanese/fallback cases that this PR doesn't implement. Could we replace these with a small mocked transcription-path regression and keep the test changes scoped to this feature?

PR #2809, opened 2026-07-12
Hardware-aware inference optimizations: flash attention, hook-free KV cache, vectorized beam search

RozhinKh:
## Summary

Four hardware-aware optimizations targeting Whisper's inference path on Ampere GPUs (RTX 3090 Ti, CUDA 12.1). All changes are validated to produce byte-identical transcripts; two of the four apply equally to any GPU with PyTorch SDPA support. The hook-based path is fully preserved for backward compatibility.

---

## Optimization 1 — Flash attention in cross-attention during beam search (largest contributor)

**Problem:** Cross-attention key/value tensors are computed once per audio segment at `batch=n_audio` and shared across all beam hypotheses at `batch=n_audio * beam_size`. This batch size mismatch causes PyTorch's Flash Attention and memory-efficient SDPA kernels to silently reject the inputs and fall back to the unfused math kernel — with no warning and a significant speed penalty on every cross-attention call during decoding.

**Fix (`whisper/model.py`):** Pre-expand `k` and `v` to match `q`'s batch size before `scaled_dot_product_attention`. `torch.expand` returns a stride-0 view — no allocation, no data copy.

```python
if k.shape[0] == 1 and q.shape[0] != 1:
    k = k.expand(q.shape[0], *k.shape[1:])
    v = v.expand(q.shape[0], *v.shape[1:])
a = scaled_dot_product_attention(q, k, v, is_causal=mask is not None and n_ctx > 1)
```

**Isolated benchmark (large-v3, RTX 3090 Ti):**

| Kernel | Latency per call |
|--------|-----------------|
| Math kernel (original) | 174.65 µs |
| Flash attention (this PR) | 20.48 µs |

The condition `k.shape[0] == 1 and q.shape[0] != 1` is conservative: no effect on self-attention, greedy decoding, `beam_size=1`, or multi-audio batches. Fully backward-compatible on GPUs without flash attention support.

---

## Optimization 2 — Hook-free KV cache

**Problem:** `install_kv_cache_hooks` registers `register_forward_hook` callbacks on every key and value projection in the decoder — 128 hooks per decode step for `large-v3` (32 layers × 2 attention types × 2 projections). Each hook call carries Python dispatch overhead and prevents `torch.compile` from tracing a stable graph.

**Fix (`whisper/model.py`, `whisper/decoding.py`):** Assign a stable integer index `_kv_idx` to each `MultiHeadAttention` layer in `TextDecoder.__init__`. A new `Whisper.make_kv_cache()` method returns `{idx: [None, None]}` dicts. KV accumulation then happens directly inside `MultiHeadAttention.forward()` via list mutation — the list object identity is stable across calls, so `torch.compile` dynamo guards never fire on value identity.

The existing hook-based path is preserved behind a `_kv_idx is None` guard for any code that calls `install_kv_cache_hooks()` directly.

---

## Optimization 3 — Vectorized `BeamSearchDecoder.update()`

**Problem:** The original `update()` loop calls `.item()` twice per candidate token — once for the log-prob and once for the token id — for every beam in every audio. At `beam_size=5` that is `n_audio × 5 × 6 × 2 = 60` individual CPU-GPU synchronization points per decode step, each one stalling the GPU pipeline to transfer a single scalar.

**Fix (`whisper/decoding.py`):** Compute all top-`(beam_size+1)` candidates for every row in one batched `topk`, add cumulative scores with `unsqueeze` + broadcast, then transfer everything to CPU in a single `.tolist()`. The inner loop body is unchanged — only the data source changes from `.item()` per call to indexing pre-fetched Python lists.

```python
topk_vals, topk_idx = logprobs.topk(self.beam_size + 1, dim=-1)
new_logprobs = sum_logprobs.unsqueeze(1) + topk_vals
new_logprobs_list = new_logprobs.tolist()
topk_idx_list = topk_idx.tolist()
tokens_list = tokens.tolist()
```

`rearrange_kv_cache` is updated in parallel: `source_indices` is converted to a tensor once per step (instead of once per layer), and cross-attention entries (odd indices) are skipped since their batch dimension is `n_audio`, not `beam_size`.

---

## Optimization 4 — In-place mel spectrogram ops

**Problem:** The power spectrum computation calls `abs()` which internally computes `sqrt(real² + imag²)`, then immediately squares the result — the square root is computed and discarded. The log-mel normalization creates three intermediate tensors via `torch.clamp`, `torch.maximum`, and arithmetic on temporaries.

**Fix (`whisper/audio.py`):** Compute `real² + imag²` directly, eliminating the sqrt. Apply all log-mel normalization in-place (`clamp_`, `log10_`, `add_`, `div_`), reducing intermediate allocations from 4 to 1 per audio chunk.

```python
s = stft[..., :-1]
magnitudes = s.real.square().add_(s.imag.square())
...
log_spec = mel_spec.clamp_(min=1e-10).log10_()
log_spec.clamp_(min=log_spec.max().item() - 8.0)
log_spec.add_(4.0).div_(4.0)
```

---

## End-to-end results

Measured on `large-v3`, `beam_size=5`, fp16 (RTX 3090 Ti, CUDA 12.1, PyTorch 2.5.1). Evaluation: LibriSpeech test-clean, 200-sample stride-13 sweep.

| Metric | Baseline | This PR | Delta |
|--------|----------|---------|-------|
| RTF (mean) | 0.1677 | 0.1313 | **−21.7%** |
| WER | 7.002% | 7.002% | **0.00%** |

Transcripts are byte-identical before and after. Verified across 7 evaluation scenarios covering clean speech, noisy conditions (SNR 10/20 dB), telephone audio, and long-form content, with end-to-end speedups ranging from 7% to 36% depending on audio characteristics. Concurrent correctness verified with 6 simultaneous requests (all byte-identical).

---

## What this does not change

- Public API: no changes to `transcribe()`, `decode()`, `detect_language()`, or any function signatures
- Hook-based path: `install_kv_cache_hooks()` still works and is still called for any `MultiHeadAttention` layer that has `_kv_idx = None`
- Non-SDPA path: math kernel fallback in `qkv_attention` is unchanged
- Dependencies: no new packages

### DIFF
--- a/scripts/find_optimal_config.py
+++ b/scripts/find_optimal_config.py
@@ -1,0 +1,119 @@
+"""
+Grid search over beam_size, compute_type, and torch.compile to find the fastest
+configuration for Whisper large-v3 on RTX 3090 that keeps WER within tolerance.
+
+Usage:
+    python scripts/find_optimal_config.py --audio path/to/test.wav [--reference "ground truth text"]
+"""
+
+import argparse
+import time
+import warnings
+
+import torch
+import whisper
+
+CONFIGS = [
+    {"compute_type": "float16", "beam_size": 1, "use_compile": False},
+    {"compute_type": "float16", "beam_size": 3, "use_compile": False},
+    {"compute_type": "float16", "beam_size": 5, "use_compile": False},
+    {"compute_type": "float16", "beam_size": 1, "use_compile": True},
+    {"compute_type": "float16", "beam_size": 3, "use_compile": True},
+    {"compute_type": "float32", "beam_size": 1, "use_compile": False},
+    {"compute_type": "float32", "beam_size": 5, "use_compile": False},
+]
+
+
+def wer(reference: str, hypothesis: str) -> float:
+    ref_words = reference.lower().split()
+    hyp_words = hypothesis.lower().split()
+    # simple dynamic programming WER
+    d = [[0] * (len(hyp_words) + 1) for _ in range(len(ref_words) + 1)]
+    for i in range(len(ref_words) + 1):
+        d[i][0] = i
+    for j in range(len(hyp_words) + 1):
+        d[0][j] = j
+    for i in range(1, len(ref_words) + 1):
+        for j in range(1, len(hyp_words) + 1):
+            cost = 0 if ref_words[i - 1] == hyp_words[j - 1] else 1
+            d[i][j] = min(d[i - 1][j] + 1, d[i][j - 1] + 1, d[i - 1][j - 1] + cost)
+    return d[len(ref_words)][len(hyp_words)] / max(len(ref_words), 1)
+
+
+def benchmark_config(audio_path: str, config: dict, n_warmup: int = 1, n_runs: int = 3):
+    model = whisper.load_model(
+        "large-v3",
+        compute_type=config["compute_type"],
+        use_compile=config["use_compile"],
+    )
+
+    # warm-up
+    for _ in range(n_warmup):
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore")
+            model.transcribe(audio_path, beam_size=config["beam_size"], fp16=(config["compute_type"] == "float16"))
+
+    # timed runs
+    times = []
+    result_text = ""
+    for _ in range(n_runs):
+        torch.cuda.synchronize()
+        t0 = time.perf_counter()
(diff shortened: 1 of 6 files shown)


RozhinKh:
Withdrawing this PR for now -- needs internal team review/approval before it goes up for review here. Will reopen or resubmit once that's done.

PR #2811, opened 2026-07-13
feat: add transcription progress callback

Samielakkad:
## What changed

Adds an explicit `progress_callback` keyword to `whisper.transcribe`.

The callback receives a monotonic float from `0.0` to `1.0` after each decoded segment. It is kept separate from `**decode_options`, so it cannot be accidentally forwarded to `DecodingOptions`.

Also adds:

- a README usage example
- integration assertions in `tests/test_transcribe.py` for callback invocation, monotonicity, and completion

## Why

Discussion #2810 reports that users reasonably try to pass a progress callback, but the unknown `progress` keyword is forwarded into `DecodingOptions` and raises:

```
TypeError: DecodingOptions.__init__() got an unexpected keyword argument 'progress'
```

This adds a small backwards-compatible API for applications that need progress without monkey-patching the internal `tqdm` implementation. Existing callers are unchanged.

## Validation

- `python -m compileall -q whisper/transcribe.py tests/test_transcribe.py`
- AST check confirms the new keyword-only parameter
- `git diff --check`
- README callback snippet reviewed

The full model integration test was not runnable in this checkout because PyTorch is not installed; CI can run the existing model matrix.

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,6 +122,16 @@ result = model.transcribe("audio.mp3")
 print(result["text"])
 ```
 
+To receive transcription progress in an application, pass a callback that accepts a float from `0` to `1`:
+
+```python
+def on_progress(progress: float):
+    print(f"Transcription: {progress:.0%}")
+
+
+result = model.transcribe("audio.mp3", progress_callback=on_progress)
+```
+
 Internally, the `transcribe()` method reads the entire file and processes the audio with a sliding 30-second window, performing autoregressive sequence-to-sequence predictions on each window.
 
 Below is an example usage of `whisper.detect_language()` and `whisper.decode()` which provide lower-level access to the model.
--- a/tests/test_transcribe.py
+++ b/tests/test_transcribe.py
@@ -12,11 +12,19 @@ def test_transcribe(model_name: str):
     device = "cuda" if torch.cuda.is_available() else "cpu"
     model = whisper.load_model(model_name).to(device)
     audio_path = os.path.join(os.path.dirname(__file__), "jfk.flac")
+    progress = []
 
     language = "en" if model_name.endswith(".en") else None
     result = model.transcribe(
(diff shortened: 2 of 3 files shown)


Samielakkad:
@rabbitxyz2026 you reported the progress failure in #2810. I opened [PR #2811](https://github.com/openai/whisper/pull/2811) with an explicit progress_callback(progress: float) API, README usage, and integration assertions. If you can test it against your Windows/CUDA setup and report whether the callback semantics fit your application, I’ll incorporate any concrete changes with your authorship credited.

Samielakkad:
Closing this as a duplicate after finding the earlier progress-callback implementations in #1025, #2469, #2625, and especially the newer tested #2797. Keeping another competing API PR open would add noise rather than help maintainers converge on one design.

rabbitxyz2026 (commented):
update progress

rabbitxyz2026:
> @rabbitxyz2026 you reported the progress failure in #2810. I opened [PR #2811](https://github.com/openai/whisper/pull/2811) with an explicit progress_callback(progress: float) API, README usage, and integration assertions. If you can test it against your Windows/CUDA setup and report whether the callback semantics fit your application, I’ll incorporate any concrete changes with your authorship credited.

Thanks! I tested your PR on my Windows + CUDA setup, and it works well.

The `progress_callback(progress: float)` API behaves exactly as I expected, and it integrates nicely with my application. I was able to use it to send real-time transcription progress over WebSocket without modifying `transcribe.py`.

I don't have any further suggestions at the moment. The current callback semantics fit my use case perfectly.

I have one question, though. 
Is there currently a supported way to cancel an ongoing `transcribe()` operation? For example, I'd like to allow a user to stop transcription immediately from the UI while it's still running. Is this possible, or would it require modifying the transcription loop internally?

Thanks again for implementing this feature!

PR #2823, opened 2026-08-01
Fix: carry_initial_prompt overflows the decoder prompt budget

Yigtwxx:
## Problem

`carry_initial_prompt=True` exists so that a long `initial_prompt` (a glossary, a
list of proper nouns) keeps reaching the decoder on every sliding window. When
the prompt is long enough to fill the decoder's prompt budget, it does the
opposite: from the second window onwards the `initial_prompt` is dropped
entirely and replaced by carried-over transcript.

## Root cause

The prompt budget is `n_text_ctx // 2 - 1`, which is 223 tokens for the released
checkpoints (`n_text_ctx = 448`). `remaining_prompt_length` tracks how much of
that budget is left once `initial_prompt` has taken its share:

```python
remaining_prompt_length = model.dims.n_text_ctx // 2 - 1
if initial_prompt is not None:
    initial_prompt_tokens = tokenizer.encode(" " + initial_prompt.strip())
    all_tokens.extend(initial_prompt_tokens)
    remaining_prompt_length -= len(initial_prompt_tokens)   # may reach 0 or go negative
```

```python
if carry_initial_prompt:
    nignored = max(len(initial_prompt_tokens), prompt_reset_since)
    remaining_prompt = all_tokens[nignored:][-remaining_prompt_length:]
    decode_options["prompt"] = initial_prompt_tokens + remaining_prompt
```

Once `initial_prompt` is 223 tokens or longer, `remaining_prompt_length` is zero
or negative and the slice inverts:

```python
>>> tokens = [1, 2, 3, 4, 5]
>>> tokens[-2:]     # room for 2 more tokens
[4, 5]
>>> tokens[-0:]     # no room left - returns everything
[1, 2, 3, 4, 5]
>>> tokens[-(-2):]  # negative - drops from the front instead
[3, 4, 5]
```

So instead of carrying nothing, the window carries the entire accumulated
transcript. `DecodingTask._get_initial_tokens` then trims the combined prompt
from the left:

```python
prompt_tokens[-(self.n_ctx // 2 - 1) :]
```

Left-trimming removes the leading tokens, which are exactly the
`initial_prompt` tokens the option is meant to preserve.

## Fix

Clamp the remaining length at zero, and slice explicitly so the empty case
stays empty.

## Evidence

`tiny.en` on CPU, `tests/jfk.flac` tiled to ~66 s (3 windows), `temperature=0`
and `condition_on_previous_text=True` so the carry-over path runs on every
window. `DecodingTask.__init__` was wrapped to record `options.prompt`.

**`initial_prompt` of exactly 223 tokens** (budget exactly full, so nothing may
be carried):

| window | prompt length | tokens over budget | `initial_prompt` reaches the decoder |
| --- | --- | --- | --- |
| 0 | 223 → 223 | 0 → 0 | yes → yes |
| 1 | 284 → 223 | 61 → 0 | **no → yes** |
| 2 | 367 → 223 | 144 → 0 | **no → yes** |

**`initial_prompt` of 14 tokens** (room to carry, i.e. every existing use of the
option):

| window | prompt length before | prompt length after |
| --- | --- | --- |
| 0 | 14 | 14 |
| 1 | 76 | 76 |
| 2 | 134 | 134 |

Per-window prompt token ids are identical before and after, and the transcript
is byte-identical (`sha256 861f8b56223f5e5124744caacf57f1f87dc4b37a10472cecf9a5ab6abf355bc5`
in both runs).

The repository's own workflow was run on this branch before opening the PR and
is green across the whole matrix, `pre-commit` included:
https://github.com/Yigtwxx/whisper/actions/runs/30698478749

## Compatibility

No API, CLI, or default behavior change. The only inputs affected are those
where `remaining_prompt_length` was already zero or negative, i.e. an
`initial_prompt` of at least `n_text_ctx // 2 - 1` tokens combined with
`carry_initial_prompt=True`. Every shorter prompt takes the same code path as
before and produces identical output.

### DIFF
--- a/tests/test_transcribe.py
+++ b/tests/test_transcribe.py
@@ -1,5 +1,6 @@
 import os
 
+import numpy as np
 import pytest
 import torch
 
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
The code fix is small, but this really needs a committed regression for the exact boundary it changes. `remaining_prompt_length == 0` is the bug here because `tokens[-0:]` becomes the whole list; without a test at `n_text_ctx // 2 - 1` tokens it’s easy for a later refactor to reintroduce the same behavior. Could we add a focused test asserting that a full-budget initial prompt is preserved on the second window and no previous-text tokens are appended?

Yigtwxx:
Added in 34e2b00: `test_carry_initial_prompt_with_full_prompt_budget` in `tests/test_transcribe.py`.

It builds an `initial_prompt` of exactly `n_text_ctx // 2 - 1` (223) tokens, runs `transcribe()` over two 30-second windows with a model-free stub that records each window's `options.prompt`, and asserts the second window's prompt equals `initial_prompt_tokens` exactly, i.e. no previous-text tokens are appended. No model weights are loaded, so it runs in a few seconds.

Verified it fails against the current `main` (`prompts[1]` contains the first window's text tokens) and passes with the fix.

PR #2828, opened 2026-08-11
Fix suppress_tokens handling for iterables, None, and shared lists

nileshpatil6:
`DecodingOptions.suppress_tokens` is typed as `Optional[Union[str, Iterable[int]]]`, but `_get_suppress_tokens` only actually handles two of those cases cleanly. Three problems, all in the same few lines:

1. Passing any iterable that isn't a list crashes. `suppress_tokens=(5,)` hits `assert isinstance(suppress_tokens, list)` and dies with an AssertionError, even though the annotation says any iterable of ints is fine. A tuple that happens to contain -1 sneaks through via a different branch, which makes the failure look random.

2. The `None` check is dead code. `if -1 in suppress_tokens` runs before the `elif suppress_tokens is None` branch, so an explicit `None` would raise `TypeError: argument of type 'NoneType' is not iterable` on the membership test before ever reaching the branch that's meant to handle it. It's currently masked by the truthiness guard at the call site, but the function can't actually handle the value its own signature advertises.

3. When a list without -1 is passed, the later `extend()` calls append the special tokens to the caller's list in place. Decode twice with the same options object and your suppress list quietly grows with sot/translate/transcribe ids each time.

The fix normalizes the input up front: `None` becomes an empty list, strings are parsed (an empty string now parses to an empty list instead of crashing on `int("")`, matching the comment that was already there), and anything else goes through `list()`, which both accepts arbitrary iterables and copies, so the caller's list is never touched. The -1 expansion and everything after it are unchanged, and the default `"-1"` path behaves exactly as before.

Added `tests/test_decoding.py` with unit tests for the tuple/iterator/set cases, the no-mutation guarantee, and the None/empty/default behaviors. They run without a model download. Before the fix, 6 of the 8 fail; after, all pass, along with the existing tokenizer tests. Also smoke-tested a real `transcribe()` of tests/jfk.flac with the tiny model on CPU: `suppress_tokens=(5,)` used to crash and now transcribes normally, and a passed-in list is unchanged after two runs.

black and isort clean on both files.

### DIFF
--- a/tests/test_decoding.py
+++ b/tests/test_decoding.py
@@ -1,0 +1,47 @@
+from types import SimpleNamespace
+
+import pytest
+
+from whisper.decoding import DecodingOptions, DecodingTask
+from whisper.tokenizer import get_tokenizer
+
+
+def get_suppress_tokens(**kwargs):
+    tokenizer = get_tokenizer(multilingual=False)
+    task = SimpleNamespace(options=DecodingOptions(**kwargs), tokenizer=tokenizer)
+    return DecodingTask._get_suppress_tokens(task), tokenizer
+
+
+@pytest.mark.parametrize("value", [(5,), iter([5]), {5}])
+def test_suppress_tokens_accepts_any_iterable(value):
+    from_iterable, _ = get_suppress_tokens(suppress_tokens=value)
+    from_list, _ = get_suppress_tokens(suppress_tokens=[5])
+    assert from_iterable == from_list
+    assert 5 in from_iterable
+
+
+def test_suppress_tokens_does_not_mutate_the_options_list():
+    tokens = [5]
+    get_suppress_tokens(suppress_tokens=tokens)
+    assert tokens == [5]
+
+
+@pytest.mark.parametrize("value", [None, "", []])
+def test_suppress_tokens_empty_means_only_special_tokens(value):
+    suppress, tokenizer = get_suppress_tokens(suppress_tokens=value)
+    specials = {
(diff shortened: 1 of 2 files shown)


sylvesterkaczmarek (commented):
One regression in the string path: filtering every empty component makes malformed lists silently valid. `"1,,2"`, `",1"` and `"1,"` used to raise on `int("")`; this turns them into `[1, 2]` / `[1]`. If the intended change is just the documented empty-string case, could we special-case `suppress_tokens == ""` and keep strict parsing for non-empty strings? A test for an empty component would pin it.

nileshpatil6:
Good catch, that was a real regression. Filtering every empty component made malformed lists silently valid:

```
input     before this fix   after
'1,,2'    [1, 2]            ValueError
',1'      [1]               ValueError
'1,'      [1]               ValueError
''        []                []
'1,2'     [1, 2]            [1, 2]
```

Went with your suggestion and special-cased only the empty string, so any non-empty string is parsed strictly as before:

```python
suppress_tokens = (
    [int(t) for t in suppress_tokens.split(",")] if suppress_tokens else []
)
```

Added a parametrised test covering `"1,,2"`, `",1"`, `"1,"` and `"1,x"`, all asserting `ValueError`.

One note on verification: I could not run the suite locally because importing whisper fails in my environment with "Numba needs NumPy 2.3 or less. Got NumPy 2.4", which is unrelated to this change. I verified the parsing behaviour in isolation against both the old and new expressions, per the table above, and CI should cover the rest.

PR #2831, opened 2026-08-19
Fix float64 audio mel filter dtype mismatch

tandede:
## Summary

`log_mel_spectrogram()` accepts NumPy arrays and tensors, but its cached Mel filters were always `float32`. A `float64` waveform therefore produced a matrix multiplication dtype error instead of a spectrogram.

This converts the filter bank to the STFT magnitude dtype before projection. Existing `float16` and `float32` behavior is unchanged, while double-precision waveforms now produce finite Mel spectrograms. The regression test uses the repository's JFK audio and compares the `float64` result with the existing `float32` path.

Before:

```text
RuntimeError: expected m1 and m2 to have the same dtype, but got: float != double
```

After:

```text
dtype=torch.float64 shape=(80, 1100) finite=True
max_abs_difference=0.00001254
```

## Testing

- `pytest --durations=0 -q -k 'not test_transcribe or test_transcribe[tiny] or test_transcribe[tiny.en]' -m 'not requires_cuda'` (19 passed)
- `pre-commit run --all-files`

### DIFF
--- a/tests/test_audio.py
+++ b/tests/test_audio.py
@@ -14,6 +14,8 @@ def test_audio():
 
     mel_from_audio = log_mel_spectrogram(audio)
     mel_from_file = log_mel_spectrogram(audio_path)
+    mel_from_float64 = log_mel_spectrogram(audio.astype(np.float64))
 
     assert np.allclose(mel_from_audio, mel_from_file)
+    assert np.allclose(mel_from_audio, mel_from_float64, rtol=1e-4, atol=1e-4)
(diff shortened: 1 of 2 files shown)


tandede:
Hi maintainers, I wanted to follow up on this PR in case it was overlooked. It fixes the dtype mismatch encountered when float64 audio is processed with mel filters and is ready for review. I’m happy to address any feedback or add further validation if needed. Thank you!

